PR #30, opened 2014-12-16
RequestFactory: invalid byte sequences in $_SERVER['REQUEST_URI']

JanTvrdik:
The goal is to not simply ignore invalid byte sequences in URL, but throw Exception or return NULL. Current Nette behavior is unwanted, e.g. these URLs should not be accepted at all:

```
http://doc.nette.org/en/2.2/%C0%C0%C0%C0%C0templating
https://www.rohlik.cz/stranka/%C0kontakt
```

---

How would you prefer to handle it? Exception or by returning NULL. Both break the RequestFactory API. I would prefer the exception, e. g. `Nette\Http\InvalidRequestException`. Do you like the name?

cc @fprochazka

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -100,7 +100,10 @@ class RequestFactory extends Nette\Object
 
 		// normalized url
 		$url->canonicalize();
-		$url->setPath(Strings::fixEncoding($url->getPath()));
+
+		if (preg_match(self::NONCHARS, $url->getPath()) || preg_last_error()) {
+			throw new InvalidRequestException(); // TODO!
+		}
 
 		// detect script path
 		if (isset($_SERVER['SCRIPT_NAME'])) {


JanTvrdik:
Self note – Symfony deprecated scopes in 3.0, however I still think that scopes (*) can provide a good solution to this problem. Not sure though how much complexity would that bring.

(*) in general you have a directed acyclic graph of containers – services from container A can depend on services from container B only if there exist path from container A to container B. To handle the request problem you would have a *root container* and a *request container*. The request container would contain `Http\Request` instance and services depending on this instance (e.g. routers or link generator).

This makes e.g. testing presenters depending on `Http\Request` easier as you would no longer need to rebuild the entire container for each `Presenter::run` call, only the much smaller request container would need to be thrown away and recreated.

dg:
Related https://github.com/nette/routing/issues/6

PR #42, opened 2015-01-03
RequestFactory: broken script path detection due to removed canonicalize()

JanTvrdik:
Are you aware of this?


dg:
What about `$path = Url::unescape($tmp[0], '%/?#');` here https://github.com/JanTvrdik/nette-http/blob/e3a05427781848407e94f986d690acdabac22209/src/Http/RequestFactory.php#L87?

Call canonicalize() is not solution, because file path can contain for example `~`, which is escaped.

PR #43, opened 2015-01-06
Url: changed to be immutable

stekycz:
This is absolutely BC break. However there is not much other options. There could be another class `ImmutableScriptUrl` however I think it is not a good practice.

Immutability is always BC break in my opinion. The only solution is described above but it uses inheritance and immutability is only on the last class. It could be done as same as `DateTimeInterface` but you probably know how it ended in many applications :-) It could be also done as a composition but it has a disadvantage in the same way as `DateTimeInterface` - it cannot be used in all places where current `Url` is used.

### DIFF
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -62,11 +62,11 @@ class Request extends Nette\Object implements IRequest
 	public function __construct(UrlScript $url, $query = NULL, $post = NULL, $files = NULL, $cookies = NULL,
 		$headers = NULL, $method = NULL, $remoteAddress = NULL, $remoteHost = NULL, $rawBodyCallback = NULL)
 	{
-		$this->url = $url;
 		if ($query !== NULL) {
 			trigger_error('Nette\Http\Request::__construct(): parameter $query is deprecated.', E_USER_DEPRECATED);
-			$url->setQuery($query);
+			$url = $url->setQuery($query);
 		}
+		$this->url = $url;
 		$this->post = (array) $post;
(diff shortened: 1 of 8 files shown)


JanTvrdik:
This seems terrible. If we were to choose to have immutable `Url`, this is not the way to do it. What I would suggest is to split the `Url` class into immutable (not having in mutable setters) `Url` and mutable `UrlBuilder` (similar to `StringBuilder` in C#). However I still think that we don't need to enforce immutability and that current `Url` behavior is fine.

fprochazka:
For the lack of time, only short response: I don't like it. I would have liked it if it were in Nette from the beggining. But this is a massive BC Break and pretty fucking ugly one, because you won't be able to tell anything is broken.

dg:
As I [said](https://github.com/nette/http/commit/29668f68fa7e2452511b7e7522a470e5a8d6c760#commitcomment-9173502) it is hard to do changes without BC break. I cannot accept this, sry.

stekycz:
This is only quick reaction to https://github.com/nette/http/commit/29668f68fa7e2452511b7e7522a470e5a8d6c760 I am not saying it is perfect. However immutable `Url` with its builder is what I am talking about. However it also cannot be used at the same place.

@dg I think it is impossible now :-) The only immutability with complete back compatibility can be done using inheritance (to be able to use it where `Url` is used) or using builder as suggested by @JanTvrdik. However both solutions are as same partial as using `clone` for `Url` in your commit.

PR #45, opened 2015-01-18
UrlScript: added getAppPath()

JanTvrdik:
When implementing custom router you usually need this method. The algorithm was taken from https://github.com/nextras/static-router/blob/d97be13214443e03cc5efb849adc69d9ae0a2a5d/src/StaticRouter.php#L49 and it should be equivalent what Route [is doing](https://github.com/nette/application/blob/2d1e6b9880a3b375ce65df57ca3715850b6c0ecc/src/Application/Routers/Route.php#L162-L166) (just micro-optimized). Although I think I made some assumption which I'm not sure if they are generally true. However the idea is important, the algorithm could be polished.

Thoughts?

### DIFF
--- a/src/Http/UrlScript.php
+++ b/src/Http/UrlScript.php
@@ -26,6 +26,7 @@ use Nette;
  * @author     David Grudl
  *
  * @property   string $scriptPath
+ * @property-read string $appPath
  * @property-read string $pathInfo
  */
 class UrlScript extends Url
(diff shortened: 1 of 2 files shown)


dg:
How does it differ from getPathInfo()?

JanTvrdik:
Added a test which explains the difference. See also https://github.com/nextras/static-router/commit/5a137d9a6ff1a8871224ba2710864a3e19933638

dg:
Hmm, what about name? `getAppPath()` sound that it returns the exactly other part of path, or isn't it?

JanTvrdik:
Well I guess it is a but ambiguous. The idea is the router basically transforms *appPath*  to *appRequest*.

JanTvrdik:
This is still a very good idea. Suggestions for better method name are appreciated.

JanTvrdik:
I still need this and I still don't have a better name.

milo:
Something like variablePath or variableUrl? It is the part of url which is changing.

f3l1x:
Alternatives:
- appPath (8/10)
- variablePath (5/10)
- slugPath (2/10)

It might be helpful in the future. Any progress here?

dg:
I will use relativePath.

PR #49, opened 2015-02-05
Url: isEqual does not consider default ports

JanTvrdik:


JanTvrdik:
Why is Travis result not visible?

dg:
I noticed that about three days ago...

PR #50, opened 2015-02-10
Session: fetch session cookie from Http\IRequest

fprochazka:
The `nette-browser` is also fetched from `Http\IRequest`, so the `session_name()` should probably be too.

For example I might be writing my own little http server using [react.php](https://github.com/reactphp/http) that accepts requests and processes them through sockets and is started from CLI. This would mean that I have to initialize the `$_COOKIES` variable instead of passing it in a nice OO interface through manually created `Http\Request`.

The fetching of the cookie alone would help a lot. The setting of the cookie to the global variable is a paranoid safety-net, for cases where something depends on the variable existing other than `Nette\Http\Session`. If you have a problem with modifying the global variable, I can remove it. But I'd rather keep it there.

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -87,8 +87,10 @@ class Session extends Nette\Object
 
 		$this->configure($this->options);
 
-		$id = & $_COOKIE[session_name()];
-		if (!is_string($id) || !preg_match('#^[0-9a-zA-Z,-]{22,128}\z#i', $id)) {
+		$id = $this->request->getCookie(session_name());
+		if (is_string($id) && preg_match('#^[0-9a-zA-Z,-]{22,128}\z#i', $id)) {
+			session_id($id);
+		} else {
 			unset($_COOKIE[session_name()]);
 		}
 
--- a/tests/Http/Session.id.phpt
+++ b/tests/Http/Session.id.phpt
@@ -1,0 +1,31 @@
+<?php
+
+/**
+ * Test: Nette\Http\Session accepts cookie from Http\IRequest
+ */
+
+use Nette\Http,
+	Nette\Http\Session,
+	Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+$_COOKIE['PHPSESSID'] = $leet = md5(1337);
+
+// create fake session
+$cookies = array('PHPSESSID' => $sessionId = md5(1), 'nette-browser' => $B = substr(md5(2), 0, 10));
+file_put_contents(TEMP_DIR . '/sess_' . $sessionId, sprintf('__NF|a:3:{s:4:"Time";i:%s;s:1:"B";s:10:"%s";s:4:"DATA";a:1:{s:4:"temp";a:1:{s:5:"value";s:3:"yes";}}}', time() - 1000, $B));
+
+$session = new Session(new Http\Request(new Http\UrlScript('http://nette.org'), NULL, array(), array(), $cookies), new Http\Response());
+
+$session->start();
+Assert::same($sessionId, $session->getId());
+Assert::same(array('PHPSESSID' => $leet), $_COOKIE);
+
+Assert::same('yes', $session->getSection('temp')->value);
+$session->close();
+
+// session was not regenerated
+Assert::true(file_exists(TEMP_DIR . '/sess_' . $sessionId));
+Assert::count(1, glob(TEMP_DIR . '/sess_*'));
--- a/tests/Http/Session.regenerate-empty-session.phpt
+++ b/tests/Http/Session.regenerate-empty-session.phpt
@@ -1,0 +1,29 @@
+<?php
+
+/**
+ * Test: Nette\Http\Session::regenerateId() regenerate empty session
+ */
+
+use Nette\Http,
+	Nette\Http\Session,
+	Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
+// create fake session
+$cookies = array('PHPSESSID' => $sessionId = md5(3), 'nette-browser' => $B = substr(md5(4), 0, 10));
+file_put_contents(TEMP_DIR . '/sess_' . $sessionId, '__NF|a:1:{s:4:"DATA";a:1:{s:4:"temp";a:1:{s:5:"value";s:3:"yes";}}}');
+
+$session = new Session(new Http\Request(new Http\UrlScript('http://nette.org'), NULL, array(), array(), $cookies), new Http\Response());
+$session->start();
+Assert::same('yes', $session->getSection('temp')->value);
+
+$newSessionId = $session->getId();
+$session->close();
+
+// session was regenerated
+Assert::false(file_exists(TEMP_DIR . '/sess_' . $sessionId));
+Assert::true(file_exists(TEMP_DIR . '/sess_' . $newSessionId));
+Assert::count(1, glob(TEMP_DIR . '/sess_*'));


enumag:
What about `IResponse::setCookie()` and `IResponse::deleteCookie()`?

fprochazka:
@enumag you actually don't want to set the cookie using that method, because that method sends http header and is not setting the `$_COOKIES` variable.

enumag:
@fprochazka Oh. Ok :-)

dg:
I do not remember why it uses global variable https://github.com/nette/nette/issues/965 … So I am for removing global variable, but I am not sure if it is ok.

fprochazka:
@dg I believe it is safe :) What are our thoughts about the setting of the global variable?

dg:
Why are you doing it?

fprochazka:
@dg I was thinking I need it reliably set from my code, and it turns out I probably don't need it directly.

The problem is that when you pass the object with cookies in, the `session_start()` is not actually using the values you've provided.

```php
$_COOKIES['PHPSESSID'] = md5(2);
$cookies = array('PHPSESSID' => $sessionId = md5(1));
$session = new Session(
    new Http\Request(new Http\UrlScript('http://nette.org'), NULL, array(), array(), $cookies),
    new Http\Response()
);
```

I think it's pretty obvious, which value is going to be used as session id. Unless the Request object actually has effect on the session, there is no reason to pass the cookies in it and read them from it.

We might wanna restore the previous global state (I see no reason for that currently), but the global has to be set, at least temporarily.

I've changed the test to reflect that.

dg:
Uff, I have absolutely no idea why I used global $_COOKIE…

fprochazka:
@dg have I made myself clear? So what's the plan? Do you wanna restore the global state? Or is it good as is and mergable?

fprochazka:
I'm actually going to rebase it into one commit, as it makes no sense separated.

fprochazka:
Done.

dg:
Oh wait, session_start takes session id from $_COOKIE?

fprochazka:
Yes, that's why I'm changing it using the value from `Http\Request`.

fprochazka:
I see now, it can be changed using [session_id](http://php.net/manual/en/function.session-id.php)

dg:
Before it is started?

fprochazka:
I have no ide, I'd have to test that.

dg:
And test whether it is readable before session is started, for validation.

fprochazka:
Fast observation
- `$_COOKIE` is default
- if you call `session_id()` before `session_start()` it takes precedence over `$_COOKIE`
- `session_id()` is not affected by `$_COOKIE`

dg:
And for reading?

fprochazka:
this works

```php
session_id($leet = md5(1337));
$_COOKIE['PHPSESSID'] = md5(1);
Assert::same($leet, session_id($leet));
```

fprochazka:
this works

```php
session_id($leet = md5(1337));
$_COOKIE['PHPSESSID'] = md5(1);
Assert::same($leet, session_id());
```

dg:
So value of cookie can be read via `session_id()`? In PHP 5.3 too?

fprochazka:
It cannot, 

> session_id() is not affected by $_COOKIE

fprochazka:
but if you don't call `session_id($value)` and start session with `session_start()` then the session id is read from `$_COOKIE` and logically, the `session_id()` then returns it's value

dg:
So `$this->request->getCookie(session_name());` is good, but `$_COOKIE[session_name()] = $id;` is unnecessary.

fprochazka:
it works like this

```php
$_COOKIE['PHPSESSID'] = $leet = md5(1337);
session_start()
Assert::same($leet, session_id());
```

..

```php
$_COOKIE['PHPSESSID'] = $leet = md5(1337);
Assert::same(NULL, session_id());

session_id($boring = md5(1))
session_start()
Assert::same($boring, session_id());
```

fprochazka:
@dg yes, I'm changing it to work with `session_id`

fprochazka:
@dg all green. I was thinking I might create more tests but I don't feel very creative right now.

dg:
What about `unset($_COOKIE[session_name()])`? Is it needed?

fprochazka:
I think it is. If you provide invalid ID the `session_start()` might try to pick it up anyway and throw error. Unseting it guarantees that new ID will be generated.

fprochazka:
Try commenting the line and running `tests/Http/Session.invalidId.phpt`

dg:
When you replace it with ie `session_is(NULL)`?

dg:
It is maybe the only one place where Nette modifies global variable.

fprochazka:
Neither `NULL` nor empty string `''` works as reset of the id. The only way would be generate new one using `Utils\Ranom`.

fprochazka:
Rebased.

dg:
You said that calling session_id() before session_start() it takes precedence over $_COOKIE. In this case there must be called session_id() to ensure it really uses checked value.

fprochazka:
So using the provided value and throwing an exception is better than ignoring it completely and generating new one?

fprochazka:
The check was introduced exactly for this reason, so people cannot generate e500 on my webside by simply fucking with cookies.

dg:
Sorry, I am 30 awake and I was looking at the wrong diff…

fprochazka:
I'm usually really dumb after beeing 30h awake :) I'd suggest you to have some sleep and look at this tomorrow :)

dg:
So the last thing: can you turn over condition? `if (is_string($id) && preg_match('#^[0-9a-zA-Z,-]{22,128}\z#i', $id)) { ...`

fprochazka:
Surely can! Done :)

dg:
Follow this:

1) $_COOKIE contains bad value and I will call session_id('bad value')
2) `preg_match('#^[0-9a-zA-Z,-]{22,128}\z#i', $id)` is FALSE
3) `unset($_COOKIE[session_name()]);`
4) session_start will use 'bad value', is it right?

fprochazka:
once you call `session_id('something')`, it doesn't matter what value has the `$_COOKIES`

fprochazka:
That's why we have to check it and set the value (or unset the `$_COOKIES`, as there is no other way of protecting it from beeing fed to `session_start()`).

fprochazka:
If you're asking if you can still break it by providing broken cookie to `Http\Request` and then calling `session_id('another broken value')` just before calling `Session->start()` then yes, you can break it that way.

fprochazka:
As you cannot reset the value once provided to `session_id`, you'd have to manually generate new unique session id to be really safe.

fprochazka:
I'm gonna try if explicitly calling `regenerate_session_id()` fixes it.

dg:
So session_id() must be called in both states.

fprochazka:
I tried generating new ID manually, but it breaks several other tests. I don't know how to reset the id.

I have no time to investigate anymore. Maybe at friday.

dg:
session_regenerate_id will not help?

fprochazka:
It has no effect before session is started.

dg:
Use static value and set `$this->regenerated = TRUE`?

fprochazka:
That breaks [tests/Http/Session.handler.phpt](https://github.com/nette/http/blob/master/tests/Http/Session.handler.phpt) and some other tests.

fprochazka:
Thank you for merging it. I was going to investigate more though :) I may have some time on weekend to look into this more.

PR #51, opened 2015-02-10
Session: fix handling exceptions thrown from SessionHandlerInterface

fprochazka:
I need to throw exception the SessionHandler.

Currently, when you throw an exception from there, the php thinks the session is started (it doesn't care about the exception) and Nette think it's not started, because the `self::$started = TRUE;` wasn't executed.

This creates a buggy state, where you cannot call `$session->start()` again, because it throws `Nette\InvalidStateException("session_start(): session is already started")`

Failing test is provided.

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -92,15 +92,18 @@ class Session extends Nette\Object
 			unset($_COOKIE[session_name()]);
 		}
 
-		// session_start returns FALSE on failure only sometimes
-		Nette\Utils\Callback::invokeSafe('session_start', array(), function($message) use (& $error) {
-			$error = $message;
-		});
+		try {
+			// session_start returns FALSE on failure only sometimes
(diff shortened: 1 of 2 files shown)


fprochazka:
As this is a bugfix of a thing that nobody could have ever used before... Is there any chance this might get into 2.3.0, so I can create a dependency in Kdyby/Redis?

dg:
Simplification?

```php
		try {
			// session_start returns FALSE on failure only sometimes
			Nette\Utils\Callback::invokeSafe('session_start', array(), function ($message) use (& $e) {
				$e = new Nette\InvalidStateException($message);
			});
		} catch (\Exception $e) {
		}

		Helpers::removeDuplicateCookies();
		if (!empty($e)) {
			@session_write_close(); // this is needed
			throw $e; // todo: should this be wrapped in Nette\InvalidStateException ?
		}
```

fprochazka:
I like that! Updated.

fprochazka:
Awesome! :)

PR #52, opened 2015-02-11
RequestFactory: reverted that script path detection is case-sensitive

dg:
This is very annoying on Windows. Either it must give a meaningful exception message, or be case insensitive.

This reverts commit 06cc6f48b73a2ab2a3c902e4d869eeef4aa810ce

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -90,10 +90,11 @@ class RequestFactory extends Nette\Object
 		$url->setQuery(isset($tmp[1]) ? $tmp[1] : '');
 
 		// detect script path
-		$script = isset($_SERVER['SCRIPT_NAME']) ? $_SERVER['SCRIPT_NAME'] : '';
-		if ($path !== $script) {
-			$max = min(strlen($path), strlen($script));
-			for ($i = 0; $i < $max && $path[$i] === $script[$i]; $i++);
+		$lpath = strtolower($path);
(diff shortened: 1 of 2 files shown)


JanTvrdik:
I don't understand how you can end up on the wrong URL anyway. Do you write the URL by hand? I use http://larsjung.de/h5ai/ on localhost ([screen](http://screencast.com/t/TURXLpjY)) and I use clicking on links when visiting project for the first time or browser URL bar completion which contains correct URLs.

dg:
It doesn't matter. The problem is that in the root the web works, once you click on the link, you will get a 404. This could become a big issue for many users.

dg:
(Btw for opening web sites I am using file association `"C:\Program Files\Internet Explorer\iexplore.exe" "http://localhost/%1"`)

JanTvrdik:
> It doesn't matter.

It does matter. If they do not end up on the *wrong URL*, everything is fine. If we choose to revert the fix, it should certainly be done for Windows only. Doing so on Linux would be technically wrong. Doing different thing on Windows and on Linux is also not ideal.

dg:
It doesn't matter, it just happens sometimes. And web suddenly starts to behave chaotic.

dg:
> Doing so on Linux would be technically wrong.

@JanTvrdik  why?

JanTvrdik:
The detected path would be wrong, wouldn't it? You changed a test to match Windows behavior which is not correct on Linux. Or am I wrong?

dg:
I don't understand… When it will detect wrong path on linux?

PR #53, opened 2015-02-13
Request: added ::getReferrer (RFC 1945 misspelling)

hranicka:
I see, Nette\Http wraps HTTP headers and for some of them offers direct accessor methods.

But header *Referer* is misspelled (but accepted in RFC 1945) and I think we should hide this typo in object-oriented wrapper.
Instead of `Request::getReferer` we should offer correct word `Request::getReferrer`.

What do you think about this?

PHP itself uses HTTP_REFERER as server variable, header is really *referer* and nothing else.
I know it. But shouldn't we use correct accessor name (without misspelling)?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -51,7 +51,7 @@ echo $httpRequest->getRemoteHost();    // and its DNS translation
 What URL the user came from? Returned as [Nette\Http\Url |urls] object.
 
 ```php
-echo $httpRequest->getReferer()->host;
+echo $httpRequest->getReferrer()->host;
 ```
 
 Request parameters:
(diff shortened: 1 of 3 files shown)


dg:
:-1:

fprochazka:
I always get confused, because the work is misspelled, but I don't think its good to fix it like this. After all, it's in the RFC and everybody uses it.

hranicka:
Thx for your feedback. It's shortcut to header and header is mispelled `referer` so we should respect it. Crazy 90s.

dg:
Intentionnaly misspeled.

PR #54, opened 2015-02-18
Url::getQuery: PHP version condition fixed

hranicka:
It was a typo in condition for PHP version checking.
"Else" condition was never been called.

Result of the method should be unchanged after the PR.


fprochazka:
Nice catch  !

dg:
Thx

PR #55, opened 2015-02-19
HttpRequest: drops non-UTF8 strings, but control characters only removes...

dg:
http://forum.nette.org/cs/22233-sanitizace-dat-v-httprequestu

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -19,7 +19,7 @@ use Nette,
 class RequestFactory extends Nette\Object
 {
 	/** @internal */
-	const CHARS = '#^[\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]*+\z#u';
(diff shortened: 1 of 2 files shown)


src/Http/RequestFactory.php
```diff
@@ -112,19 +112,20 @@ public function createHttpRequest()
...
 						unset($list[$key][$k]);
 
 					} elseif (is_array($v)) {
 						$list[$key][$k] = $v;
 						$list[] = & $list[$key][$k];
 
-					} elseif (!preg_match(self::CHARS, $v) || preg_last_error()) {
-						$list[$key][$k] = '';
+					} else {
+						$list[$key][$k] = (string) preg_replace('#[^' . self::CHARS . ']+#u', '', $v);
```
JanTvrdik: Have you compared performance with doing preg_match (to check if replace is needed) first?

dg: Didn't but utf-8 check is extreme fast.

Issue #57, opened 2015-02-26
Http POST with Content-type: application/json 


novotnyj:
Request post attribute is currently filled with data from $_POST array. Would it be possible to set it like this
`Nette\Utils\Json::decode(file_get_contents('php://input'), true);` when method is POST and Content-type is application/json?

fprochazka:
Definitely not by default. But an option, maybe?

JanTvrdik:
:+1: on built-in support for application/json. I'm however not sure what's the best way to do it.

novotnyj:
@fprochazka  Why do you think it shouldn't be on by default?
Or maybe other solution would be to add a new method to Request class - getJSON. It would be more general and usable for other request methods (PUT,...).

fprochazka:
I wouldn't like having it in `$post` of `Http\Request`. But I guess it wouldn't break anything.. right?
Maybe having it auto decode and set as `$post` of `Application\Request` in router layer, would be better?

JanTvrdik:
Adding `getJson()` to `Http\Request` is a "ok but conservative" way to do it. The question is whether we *want* to be smart about it. 

One option would be to add general `getBody()` which would return `getRawBody()` parsed according to current Content-Type.

enumag:
@JanTvrdik And how should we add support for custom types? Don't forget that `Http\Request` should not be a service in the future.

JanTvrdik:
@enumag Remember that `Http\Request` is only a dummy value object. The actual data parsing would be done in `Http\RequestFactory` which will be service in the future.

JanTvrdik:
Does anybody feel like writing a prototype? :smile:

novotnyj:
I have made a getJSON method draft: novotnyj/http@0e68551bc18b22c4e587180483cc3eed436511b4

PR #58, opened 2015-02-27
Request::getBody(), see #57

JanTvrdik:


src/Http/Request.php
```diff
@@ -293,7 +298,17 @@ public function getRemoteHost()
 	 */
 	public function getRawBody()
 	{
-		return $this->rawBodyCallback ? call_user_func($this->rawBodyCallback) : NULL;
+		return $this->rawBodyCallback ? call_user_func($this->rawBodyCallback, $this) : NULL;
```
fprochazka: IMHO rawbody should return rawbody, not parsed body

src/Http/Request.php
```diff
@@ -58,9 +58,14 @@ class Request extends Nette\Object implements IRequest
...
+	/** @var callable|NULL */
+	private $bodyCallback;
+
 
-	public function __construct(UrlScript $url, $query = NULL, $post = NULL, $files = NULL, $cookies = NULL,
-		$headers = NULL, $method = NULL, $remoteAddress = NULL, $remoteHost = NULL, $rawBodyCallback = NULL)
+	public function __construct(
+		UrlScript $url, $query = NULL, $post = NULL, $files = NULL, $cookies = NULL,
+		$headers = NULL, $method = NULL, $remoteAddress = NULL, $remoteHost = NULL,
+		$rawBodyCallback = NULL, $bodyCallback = NULL)
```
Majkl578: You miss the assignment for bodyCallaback.

src/Http/Request.php
JanTvrdik: ? This commit does not change `getRawBody()` result.

Majkl578:
Although it's a nice closure magic, I'm afraid it's gonna be slow, it's not even lazy... And those unnecessary body calls could be easily avoided if type check was extracted outside (like `addBodyCallback('application/json', $cb)`).

src/Http/Request.php
fprochazka: You're using the same callback, that also parses the JSON for application/json, aren't you?

fprochazka: Oh, you've added new parameter, sorry :)

JanTvrdik:
@Majkl578 That's how I originally wanted to do it, but then I though that it is not powerful enough. What if you want to make decision based on something different than Content-Type (e.g. HTTP method)? What if you want to handle all something/* types? That being said I like the simple solution `addBodyCallback('application/json', $cb)` more than the current one.

JanTvrdik:
Sidenote: it must somehow workaround the JSON decode DOS vulnerability.

JanTvrdik:
I think that this is now ready to merge.

src/Http/Request.php
```diff
@@ -289,7 +296,18 @@ public function getRemoteHost()
...
+	}
+
+
+	/**
+	 * Returns parsed content of HTTP request body.
+	 * @return mixed
+	 * @throws InvalidRequestBodyException
+	 */
+	public function getBody()
+		{
```
enumag: indentation

enumag:
@JanTvrdik What about the JSON decode DOS vulnerability? I didn't see anything that would solved it in your code (but I might have missed it of course).

JanTvrdik:
@enumag I ignored it the same way everybody in the PHP worlds ignores it. Those few that are concern may replace the body parser for JSON with sth smarter.

JanTvrdik:
BTW: PSR7 calls this method `getParsedBody()` instead of just `getBody()`. It is longer but a bit more descriptive.

src/Http/RequestFactory.php
```diff
@@ -235,7 +267,22 @@ public function createHttpRequest()
 			return file_get_contents('php://input');
 		};
 
-		return new Request($url, NULL, $post, $files, $cookies, $headers, $method, $remoteAddr, $remoteHost, $rawBodyCallback);
+		// parsed body
+		$bodyCallback = function (Request $request) use (& $body) {
+			if ($body === NULL) {
+				$contentType = $request->getHeader('Content-Type');
+				foreach ($this->bodyParsers as $parserContentType => $parser) {
+					if (stripos($contentType, $parserContentType) === 0) {
```
dg: strcasecmp is better

src/Http/RequestFactory.php
```diff
@@ -33,6 +34,25 @@ class RequestFactory
...
 	private $proxies = [];
 
+	/** @var callable[] of function (Request $request): mixed */
+	private $bodyParsers = [];
+
+
+	public function __construct()
+	{
+		$this->addBodyParser('application/x-www-form-urlencoded', function (Request $request) {
+			return $request->getPost();
```
dg: There are also files…

dg:
It's a generous solution, but I wonder if it makes sense to have parsers something other than JSON and XML.

src/Http/RequestFactory.php
JanTvrdik: You mean strncasecmp?

src/Http/RequestFactory.php
JanTvrdik: Files are not send with application/x-www-form-urlencoded mime type but with multipart/form-data.

JanTvrdik:
CSV, maybe.

src/Http/RequestFactory.php
dg: Yes. Its PHP shame that there is no startsWith function.

src/Http/RequestFactory.php
dg: Yes, you right. 

But still: it is useful?

dg:
- getBody() vs getParsedBody(): imho second one is better
- support for application/x-www-form-urlencoded: nice, but IMHO useless
- support for application/json: really useful, but doesn't cover fact that Json::decode has option FORCE_ARRAY
- support for csv: imho not useful, but again - csv has not strong format, delimiter can be `,` or `;`
- support for xml: very useful, has strong format, but - should we return SimpleXMLElement  or DOMDocument?

The question is: should be this decised/configured in RequestFactory? Should be it done on HTTP level? 

And next on: is getParsedBody() really useful when it can return anything? When it returns array, application must checked as well content-type header, whether it is JSON or CSV.

So I think that this is nice feature but not good abstraction. What about new generic function `isContentType($type)` or specific isJson() & isXml(), or even getJsonBody()?

PR #59, opened 2015-04-07
Url: fixed isEqual() for same param values & regular sorting on numbers

hrach:
I wonder if anyone has ever used this for production :) Seems to be quite buggy.


dg:
Nice :)

Majkl578:
I guess this isn't enough when any of the parameters is an array?

hrach:
There is a test for this few lines above. Take a look. Maybe is something still missing.

PR #60, opened 2015-04-27
add generic get method

niklongstone:
add a generic get  method to the Request class.
it is an alternative use of getQuery, getPost and getFile method.

- feature
- issues - none
- documentation - needed: nette/web-content#285
- BC break - no

### DIFF
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -14,6 +14,7 @@ use Nette;
  * HttpRequest provides access scheme for request sent via HTTP.
  *
  * @author     David Grudl
+ * @author     Nicola Pietroluongo <nik.longstone@gmail.com>
  *
  * @property-read UrlScript $url
  * @property-read array $query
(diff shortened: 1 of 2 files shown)


src/Http/Request.php
```diff
@@ -176,8 +177,34 @@ public function getCookies()
...
+    /**
+     * Provides a flexible generic get method.
+     * It follows an order of precedence: GET, POST, FILE.
+     *
+     * @param $key
+     * @param $default
+     *
+     * @return mixed
+     */
+    public function get($key, $default = NULL)
```
fprochazka: `$_REQUEST` all over again? :)

fprochazka: :-1:

JanTvrdik: It's safer that `$_REQUEST` because it does not allow cookies.

fprochazka: True :) But it still doesn't fee alright.

fprochazka: True :) But it still doesn't feel right.

Majkl578: `$_REQUEST` only includes cookies if `request_order` is misconfigured. It does not include them by default.

niklongstone: I think it worth it. 
This central method (is like a shortcut), the user won't be afraid about the HTTP verbs used in the Request, and could easily allow interface for other libraries.

dg:
:-1:

dg:
Thank you for your pull request, but I don't want to add this.

PR #64, opened 2015-06-16
Configurable name of cookie browserKey

jasin755:


JanTvrdik:
It breaks tests. You cannot use `$options` like that.

jasin755:
What is the correct way?

Unlink:
I think, you should add new field for that and sets it in `SessionExtension`.
`$options` are used for [php session variables](https://php.net/manual/en/session.configuration.php)

dg:
I think that configurable name is needless complexity, better is to use constant.

jasin755:
Occasionally are found clients who for some reason refuses to show what framework is used. There is no way (except hack) how to accommodate them.

dg:
Closed by #103

PR #66, opened 2015-06-25
Fix session_regenerate_id

lukyer:
Fixes `session_regenerate_id(): Session object destruction failed` which sometimes occures.

See same fix: https://github.com/yiisoft/yii/issues/2376

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -226,7 +226,9 @@ class Session extends Nette\Object
 			if (headers_sent($file, $line)) {
 				throw new Nette\InvalidStateException('Cannot regenerate session ID after HTTP headers have been sent' . ($file ? " (output started at $file:$line)." : '.'));
 			}
-			session_regenerate_id(TRUE);
+			if (session_id() !== '') {
+				session_regenerate_id(TRUE);
+			}
 			session_write_close();
 			$backup = $_SESSION;
 			session_start();


src/Http/Session.php
```diff
@@ -226,7 +226,7 @@ public function regenerateId()
 			if (headers_sent($file, $line)) {
 				throw new Nette\InvalidStateException('Cannot regenerate session ID after HTTP headers have been sent' . ($file ? " (output started at $file:$line)." : '.'));
 			}
-			session_regenerate_id(TRUE);
+			if (session_id() !== '') session_regenerate_id(TRUE);
```
enumag: Use multi-line if with `{ ... }`.

dg:
Can you squash it to single commit and use consistent commit message?

lukyer:
I really tried to, but you can see the mess i made using tortoisegit.
What tool do you use for rebase?

milo:
Use command line. I suppose, you have your Github fork as an `origin` remote.
```
cd http
git remote add upstream https://github.com/nette/http.git
git fetch upstream
git checkout patch-1
git rebase upstream/master
git rebase -i HEAD~20  # or how many commits you want to manage/squash/rename
# rewrite all 'pick' to 's' (as squash) and save
git push origin patch-1 --force
```

lukyer:
@milo thank you for your advice, all my complications were caused by missing `--force` flag during push

milo:
@lukyer :+1: So the last thing is the commit message to be consistent :) Something like `Session: session ID is not regenerated when not set`

lukyer:
@milo Thanks for your tutorial :) I must admit, pretty complicated way to change few letters in Nette code :) Hope it will get better with better git cli skills.

Btw: I even wasn't able to do this simple operation (squash last 2 commits) in Intellij PhpStorm :(

src/Http/Session.php
```diff
@@ -226,7 +226,9 @@ public function regenerateId()
 			if (headers_sent($file, $line)) {
 				throw new Nette\InvalidStateException('Cannot regenerate session ID after HTTP headers have been sent' . ($file ? " (output started at $file:$line)." : '.'));
 			}
-			session_regenerate_id(TRUE);
+			if (session_id() !== '') {
+				session_regenerate_id(TRUE);
+			}
 			session_write_close();
```
dg: session_write_close works when `session_id() !== ''` ?

lukyer: It seems working, but this situation is kinda rare, so if another error will show up, i will fix it here.

dg: Why it is rare?

lukyer: Probably nobody knows. Maybe PHP bug. I suspect altered session_name.
See https://github.com/yiisoft/yii/issues/2376

dg: When it is empty on line 229, it should be empty on next line 232 too, isn't it?

dg:
Do you use default or custom session handler? In case of default file handle, what is option session.save_path value?

lukyer:
Default one.
session.save_path = 'C:\xampp\tmp'
session.name = 'FDV'

PR #69, opened 2015-08-09
Request: added withUrl()

JanTvrdik:
It's often useful (e.g. in tests) to modify URL inside httpRequest. Doing this is really annoying because Request has lot of constructor parameters. The naming (`with*`) is based on PSR-7.

What do you think about this idea? Tests are currently missing.

### DIFF
--- a/src/Http/IRequest.php
+++ b/src/Http/IRequest.php
@@ -138,4 +138,11 @@ interface IRequest
 	 */
 	function getRawBody();
 
+
+	/**
+	 * Returns new IRequest instance with modified URL.
+	 * @return IRequest
+	 */
+	// public function withUrl(UrlScript $url);
+
 }
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -294,6 +294,19 @@ class Request extends Nette\Object implements IRequest
 
 
 	/**
+	 * Returns new IRequest instance with modified URL.
+	 * @return IRequest
+	 */
+	public function withUrl(UrlScript $url)
+	{
+		return new static(
+			$url, NULL, $this->post, $this->files, $this->cookies, $this->headers, $this->method,
+			$this->remoteAddress, $this->remoteHost, $this->rawBodyCallback
+		);
+	}
+
+
+	/**
 	 * Parse Accept-Language header and returns preferred language.
 	 * @param  string[] supported languages
 	 * @return string|NULL


enumag:
Can you show an example of a test where you need this? I never needed anything like it, instead it would be really benefitial for my tests if Request would not be a service (and Response as well for that matter).

fprochazka:
@JanTvrdik I like the idea, doing this every time is tedious.

JanTvrdik:
> instead it would be really benefitial for my tests if Request would not be a service

Feel free to join the [relevant discussion](https://github.com/nette/http/pull/30).

---

> Can you show an example of a test where you need this?

Sorry, I have a really bad memory. I found the code I have written (6 days ago) and the `withUrl` method would not help in that case. The reason why I brought it up today is because I was creating a draft for a router where the method would be helpful.

~~~php
class DevRouter implements Application\IRouter
{
	// *.signaly.cz -> *.signaly.l	
	// *.signaly.cz -> localhost/signaly/www/*
	public function __construct($fromMask, $toMask, Application\IRouter $innerRouter)
	{
		$this->fromMask = $fromMask;
		$this->toMask = $toMask;
		$this->innerRouter = $innerRouter;
	}

	public function match(Nette\Http\IRequest $httpRequest)
	{
		$fixedUrl = $this->transformIn($httpRequest->url);
		$fixedHttpRequest = $httpRequest->withUrl($fixedUrl); // <-- THIS IS THE LINE
		return $this->innerRouter->match($fixedHttpRequest);
	}

	public function constructUrl(Request $appRequest, Nette\Http\Url $refUrl)
	{
		$url = $this->innerRouter->constructUrl($appRequest, $refUrl);
		$fixedUrl = $this->transformOut($url);
		return $fixedUrl;
	}
}
~~~

enumag:
In my opinion you should just write a static helper method `cloneRequest($request, $url)` outside of this class.

f3l1x:
I like an idea very much. PSR-7 is IMHO stable and it has become standard part of the many (micro)frameworks. Silex, Slim, Symfony, Zend Express, Illuminate, etc.. 

Maybe we can start with withUrl and end with a whole with* methods. Like a PSR-7 said.

dg:
To avoid confusion, this has nothing to do with PSR-7, so it does not matter whether it is stable or not.

f3l1x:
@dg You are right. I had wanted to say that. Just for the discussion. :)

PR #78, opened 2015-12-29
FileUpload: Added custom file mode on move

juniwalk:
Provide custom file mode when moving file to new location.

### DIFF
--- a/src/Http/FileUpload.php
+++ b/src/Http/FileUpload.php
@@ -141,16 +141,17 @@ class FileUpload extends Nette\Object
 	/**
 	 * Move uploaded file to new location.
 	 * @param  string
+	 * @param  int
 	 * @return self
 	 */
-	public function move($dest)
+	public function move($dest, $mode = 0666)
 	{
 		@mkdir(dirname($dest), 0777, TRUE); // @ - dir may already exist
 		@unlink($dest); // @ - file may not exists
 		if (!call_user_func(is_uploaded_file($this->tmpName) ? 'move_uploaded_file' : 'rename', $this->tmpName, $dest)) {
 			throw new Nette\InvalidStateException("Unable to move uploaded file '$this->tmpName' to '$dest'.");
 		}
-		@chmod($dest, 0666); // @ - possible low permission to chmod
+		@chmod($dest, $mode); // @ - possible low permission to chmod
 		$this->tmpName = $dest;
 		return $this;
 	}


JanTvrdik:
reason?

juniwalk:
Well so I can set different mode (other than 0666) without the need to call chmod after the move since it is called in the move method.

Is it that much useless?

JanTvrdik:
But why do you need different chmod? Why you can't use 0666?

juniwalk:
Adding execute to say the least? Or when I need stricter perms (eg. readonly)?
It is not something you will you everyday but doesn't hurt to have it there, does it?

JanTvrdik:
> Adding execute to say the least?

Why would you need to add execute bit on uploaded file? That seems very dangerous.

> but doesn't hurt to have it there, does it?

That's not a reason to add it.

juniwalk:
Let's simplify it. I just saw opportunity to add something I thought could be helpful, if you disagree then hit the close button, you're the boss here.

JanTvrdik:
> I just saw opportunity to add something I thought could be helpful

Finally answer I was looking for.

> you're the boss here

No I am really not. I'm the same nobody here as you are.

---

Now that I understand your motivation we can finally talk about the feature itself. It creates inconsistency with mkdir, where the mode is still constant. What about using the `$mode` for both `mkdir` and `chmod`?

~~~php
public function move($dest, $mode = 0777)
{
	@mkdir(dirname($mode), 0777, TRUE); // @ - dir may already exist
	@unlink($dest); // @ - file may not exists
	if (!call_user_func(is_uploaded_file($this->tmpName) ? 'move_uploaded_file' : 'rename', $this->tmpName, $dest)) {
		throw new Nette\InvalidStateException("Unable to move uploaded file '$this->tmpName' to '$dest'.");
	}
	@chmod($dest, $mode & 0666); // @ - possible low permission to chmod
	$this->tmpName = $dest;
	return $this;
}	
~~~

juniwalk:
Oh, that easy? Pardon me I am very new to contributing so I took this "interrogation" little too much personally.

To be honest I am not that good with bitwise operations, what will happen with `$mode & 0666` when I set mode to `0400`for example?

JanTvrdik:
> I took this "interrogation" little too much personally

Sorry about that

> what will happen with $mode & 0666 when I set mode to 0400 for example?

It will result in 0400. The 0666 mask just removes the execute bits. However I thought about it a bit more and I don't like the solution with `$mode & 0666` very much.

juniwalk:
Yeah, safer bet would be to use two arguments, `$fileMode` and `$dirMode` but that could be confusing, right?

dg:
If I understand, you do not need this feature?

juniwalk:
@dg It is basically just sugarcoat, nothing more.

dg:
In this case I'll close it, second argument may be in future used differently.

juniwalk:
Sure, thanks.

PR #79, opened 2016-01-08
IRequest: added PATCH method constant

klimesf:
- **feature:** Added PATCH to method constants in `Nette\Http\IRequest` interface. It could be handy when designing REST APIs with Nette framework.
- **issues:** none
- **documentation:** not needed
- **BC break:** no


klimesf:
Thank you!

fprochazka:
:+1:

Issue #20, opened 2014-11-01
Implement RFC 7239 (Forwarded HTTP Extension)


JanTvrdik:
`Nette\HttpRequestFactory` should support [RFC 7239 (Forwarded HTTP Extension)](https://tools.ietf.org/html/rfc7239). It may be wise to wait for servers to implement it first.

patrickkusebauch:
About to implement it in legacy Nette 0.9 Application, where we configure our own reverse proxy. Once done there, I plan to port it for Nette\Http.

PR #81, opened 2016-01-31
RequestFactory: fix behavior behind proxy

grongor:
- bugfix
- issues - #4 
- documentation - not needed
- BC break - may break some edge-case server proxies, most servers should stay intact

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -191,9 +191,10 @@ class RequestFactory extends Nette\Object
 		$remoteHost = !empty($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
 
 		// use real client address and host if trusted proxy is used
-		$usingTrustedProxy = (bool) array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
-			return $remoteAddr !== NULL && Helpers::ipMatch($remoteAddr, $proxy);
-		});
+		$usingTrustedProxy = $remoteAddr !== NULL
+			&& (bool) array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
+				return Helpers::ipMatch($remoteAddr, $proxy);
+			});
 
 		if ($usingTrustedProxy) {
 			if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -62,11 +62,11 @@ class RequestFactory extends Nette\Object
 	{
 		// DETECTS URI, base path and script path of the request.
 		$url = new UrlScript;
-		$url->setScheme(!empty($_SERVER['HTTPS']) && strcasecmp($_SERVER['HTTPS'], 'off') ? 'https' : 'http');
+		$url->setScheme(!empty($_SERVER['HTTPS']) && strcasecmp($_SERVER['HTTPS'], 'off') !== 0 ? 'https' : 'http');
 		$url->setUser(isset($_SERVER['PHP_AUTH_USER']) ? $_SERVER['PHP_AUTH_USER'] : '');
 		$url->setPassword(isset($_SERVER['PHP_AUTH_PW']) ? $_SERVER['PHP_AUTH_PW'] : '');
 
-		// host & port
+		// host and port
 		if ((isset($_SERVER[$tmp = 'HTTP_HOST']) || isset($_SERVER[$tmp = 'SERVER_NAME']))
 			&& preg_match('#^([a-z0-9_.-]+|\[[a-f0-9:]+\])(:\d+)?\z#i', $_SERVER[$tmp], $pair)
 		) {
@@ -76,9 +76,11 @@ class RequestFactory extends Nette\Object
 			} elseif (isset($_SERVER['SERVER_PORT'])) {
 				$url->setPort((int) $_SERVER['SERVER_PORT']);
 			}
+		} else if (!empty($_SERVER['SERVER_PORT'])) {
+			$url->setPort((int) $_SERVER['SERVER_PORT']);
 		}
 
-		// path & query
+		// path and query
 		$requestUrl = isset($_SERVER['REQUEST_URI']) ? $_SERVER['REQUEST_URI'] : '/';
 		$requestUrl = Strings::replace($requestUrl, $this->urlFilters['url']);
 		$tmp = explode('?', $requestUrl, 2);
@@ -185,24 +187,33 @@ class RequestFactory extends Nette\Object
 			}
 		}
 
+		$remoteAddr = !empty($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
+		$remoteHost = !empty($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
 
-		$remoteAddr = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
-		$remoteHost = isset($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
+		// use real client address and host if trusted proxy is used
+		$usingTrustedProxy = (bool)array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
+			return $remoteAddr !== NULL && Helpers::ipMatch($remoteAddr, $proxy);
+		});
 
-		// proxy
-		foreach ($this->proxies as $proxy) {
-			if (Helpers::ipMatch($remoteAddr, $proxy)) {
-				if (isset($_SERVER['HTTP_X_FORWARDED_FOR'])) {
-					$remoteAddr = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_FOR'])));
-				}
-				if (isset($_SERVER['HTTP_X_FORWARDED_HOST'])) {
-					$remoteHost = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_HOST'])));
-				}
-				break;
+		if ($usingTrustedProxy) {
+			if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
+				$url->setScheme(strcasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') === 0 ? 'https' : 'http');
+			}
+
+			if (!empty($_SERVER['HTTP_X_FORWARDED_PORT'])) {
+				$url->setPort((int) $_SERVER['HTTP_X_FORWARDED_PORT']);
+			}
+
+			if (!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
+				$remoteAddr = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_FOR'])));
+			}
+
+			if (!empty($_SERVER['HTTP_X_FORWARDED_HOST'])) {
+				$remoteHost = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_HOST'])));
 			}
 		}
 
-
+		// method, eg. GET, PUT, ...
 		$method = isset($_SERVER['REQUEST_METHOD']) ? $_SERVER['REQUEST_METHOD'] : NULL;
 		if ($method === 'POST' && isset($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE'])
 			&& preg_match('#^[A-Z]+\z#', $_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE'])
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -76,8 +76,6 @@ class RequestFactory extends Nette\Object
 			} elseif (isset($_SERVER['SERVER_PORT'])) {
 				$url->setPort((int) $_SERVER['SERVER_PORT']);
 			}
-		} else if (!empty($_SERVER['SERVER_PORT'])) {
-			$url->setPort((int) $_SERVER['SERVER_PORT']);
 		}
 
 		// path and query
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -191,7 +191,7 @@ class RequestFactory extends Nette\Object
 		$remoteHost = !empty($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
 
 		// use real client address and host if trusted proxy is used
-		$usingTrustedProxy = (bool)array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
+		$usingTrustedProxy = (bool) array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
 			return $remoteAddr !== NULL && Helpers::ipMatch($remoteAddr, $proxy);
 		});
 
--- a/tests/Http/RequestFactory.port.phpt
+++ b/tests/Http/RequestFactory.port.phpt
@@ -1,0 +1,91 @@
+<?php
+
+use Tester\Assert;
+
+require __DIR__ . '/../bootstrap.php';
+
+/**
+ * Test of Nette\Http\RequestFactory port detection
+ */
+class RequestFactoryPortTest extends Tester\TestCase
+{
+
+	/**
+	 * @dataProvider providerCreateHttpRequest
+	 *
+	 * @param  string
+	 * @param  array
+	 */
+	public function testCreateHttpRequest($expectedPort, $server)
+	{
+		$_SERVER = $server;
+
+		$factory = new Nette\Http\RequestFactory;
+		Assert::same($expectedPort, $factory->createHttpRequest()->getUrl()->getPort());
+	}
+
+	/**
+	 * @return array
+	 */
+	public function providerCreateHttpRequest()
+	{
+		return [
+			[80, []],
+			[8080, ['HTTP_HOST' => 'localhost:8080']],
+			[8080, ['SERVER_NAME' => 'localhost:8080']],
+			[8080, ['HTTP_HOST' => 'localhost:8080', 'SERVER_PORT' => '666']],
+			[8080, ['SERVER_NAME' => 'localhost:8080', 'SERVER_PORT' => '666']],
+			[8080, ['HTTP_HOST' => 'localhost', 'SERVER_PORT' => '8080']],
+			[8080, ['SERVER_NAME' => 'localhost', 'SERVER_PORT' => '8080']],
+
+			[80, ['HTTP_X_FORWARDED_PORT' => '8080']],
+			[8080, ['HTTP_HOST' => 'localhost:8080', 'HTTP_X_FORWARDED_PORT' => '666']],
+			[8080, ['SERVER_NAME' => 'localhost:8080', 'HTTP_X_FORWARDED_PORT' => '666']],
+			[8080, ['HTTP_HOST' => 'localhost:8080', 'SERVER_PORT' => '80', 'HTTP_X_FORWARDED_PORT' => '666']],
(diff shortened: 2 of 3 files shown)


hrach:
:+1:

Majkl578:
:+1: It just took almost 2 years to implement such simple thing. :sob:

enumag:
Shouldn't we do something with HTTP_X_FORWARDED_PORT as well?

Majkl578:
@enumag I was thinking exactly the same thing. How will Nette\Http\Url and/or routers work now when https mode is enabled over port 80? Wouldn't this generate nonsense like https://example.com:80/?

grongor:
What do you think about that https://github.com/nette/http/compare/master...grongor:fix-proxy-port ?

JanTvrdik:
Has anyone considered security implication?

milo:
I was thinking about security in this way. The attacker can fake the `HTTP_X_FORWARDED_PROTO` header content. There are only two result states: `http` or `https`. These states affect the router behaviour (automatic redirections secured vs. no-flag) and URL (http:// vs https://).

Switch from `http -> https`: it can lead to application problem, but not security imho
Switch from `https -> http`: the header can be faked only outside of an SSL/TLS session, so somewhere after the proxy. But in that case you cannot trust the proxy and security does not matter.

Any other point of view?

grongor:
@JanTvrdik  @milo Yes, I was thinking about it the same way. If the proxy is able to mess with your headers then you can't trust it anyway - no matter what we put at server side. There might be a problem with misconfiguration of the proxy but we can't really predict that.

JanTvrdik:
How does other Symfony and Zend handle this?

enumag:
Symfony checks if the proxy is [trusted](https://github.com/symfony/http-foundation/blob/master/Request.php#L1925-L1928) and if so it [uses](https://github.com/symfony/http-foundation/blob/master/Request.php#L950-L958) [these](https://github.com/symfony/http-foundation/blob/master/Request.php#L1174-L1176) [headers](https://github.com/symfony/http-foundation/blob/master/Request.php#L1200-L1203) instead.

JanTvrdik:
@enumag That's what I thought. 

The current [trusted proxy check](https://github.com/nette/http/blob/8fff6ee824e2d6bc1e1a1707362205c41b4e2bad/src/Http/RequestFactory.php#L193-L194) should be extracted and used for HTTP_X_FORWARDED_PROTO as well.

enumag:
@JanTvrdik Well it seems that by default the trustedProxies array in symfony is empty. But I'm not sure if there are some added automatically  elsewhere.

Majkl578:
Symfony configures trusted proxies using kernel parameter: https://github.com/symfony/symfony/blob/83b53f4a89fa2f436ba0f8d88b0714aa534b2cd2/src/Symfony/Bundle/FrameworkBundle/FrameworkBundle.php#L53

grongor:
I added the `HTTP_X_FORWARDED_PORT` handling and taken the `$proxies` into considerations when detecting scheme and port. Is that acceptable?

src/Http/RequestFactory.php
```diff
@@ -60,25 +60,57 @@ public function setProxy($proxy)
...
+		foreach ($this->proxies as $proxy) {
+			if (Helpers::ipMatch($remoteAddr, $proxy)) {
+				$usingTrustedProxy = TRUE;
+
+				if (isset($_SERVER['HTTP_X_FORWARDED_FOR'])) {
+					$remoteAddr = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_FOR'])));
+				}
+				if (isset($_SERVER['HTTP_X_FORWARDED_HOST'])) {
+					$remoteHost = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_HOST'])));
+				}
```
enumag: I'd put these outside of the foreach.

grongor: Well, I can do that but I will have to copy the `$usingTrustedProxy === true` condition ... do you think that removing one indention is worth it?

enumag: I just think that any logic that only happens once should be outside of a cycle. Also you can skip the `=== true` part.

grongor: That makes sense, alright :) Yeah I will, I was just trying to be clear for the purpose of comment :)

src/Http/RequestFactory.php
```diff
@@ -60,25 +60,59 @@ public function setProxy($proxy)
...
+
+		if ($usingTrustedProxy) {
+			if (isset($_SERVER['HTTP_X_FORWARDED_FOR'])) {
+				$remoteAddr = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_FOR'])));
+			}
+
+			if (isset($_SERVER['HTTP_X_FORWARDED_HOST'])) {
+				$remoteHost = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_HOST'])));
+			}
+		}
```
enumag: Shouldn't this be grouped together with this code (it is about 20 lines below)? I don't see anything preventing it.

```
`if ($usingTrustedProxy && !empty($_SERVER['HTTP_X_FORWARDED_PORT'])) {
	$port = (int) $_SERVER['HTTP_X_FORWARDED_PORT'];
}
```

grongor: Yes it can be grouped together but I thought that the separation is for better readability (1. proxy, 2. scheme, 3. port, ...). If I put it together I'm gonna somehow mix these groups together ... If you think it's better I will merge these conditions.

enumag: Let's wait for other opinions. :-)

src/Http/RequestFactory.php
```diff
@@ -60,25 +60,59 @@ public function setProxy($proxy)
 	 */
 	public function createHttpRequest()
 	{
+		$remoteAddr = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
+		$remoteHost = isset($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
+
+		// use real client address and host if trusted proxy is used
+		$usingTrustedProxy = FALSE;
+		foreach ($this->proxies as $proxy) {
+			if (Helpers::ipMatch($remoteAddr, $proxy)) {
```
Majkl578: What if `$remoteAddr` is NULL?

src/Http/RequestFactory.php
Majkl578: I vote for less nesting -- merging the conditions.

src/Http/RequestFactory.php
```diff
@@ -60,25 +60,59 @@ public function setProxy($proxy)
...
+				$usingTrustedProxy = TRUE;
+				break;
+			}
+		}
```
Majkl578: This foreach could be rewritten (to avoid variable default + if + break) as such:
```php
$trustedProxies = count(array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
    return Helpers::ipMatch($remoteAddr, $proxy);
});
```

and the check below as:

```php
if ($trustedProxies) {
```

src/Http/RequestFactory.php
```diff
@@ -60,25 +60,59 @@ public function setProxy($proxy)
...
 			}
+		} else if (!empty($_SERVER['SERVER_PORT'])) {
+			$port = (int) $_SERVER['SERVER_PORT'];
+		}
+
+		if ($usingTrustedProxy && !empty($_SERVER['HTTP_X_FORWARDED_PORT'])) {
+			$port = (int) $_SERVER['HTTP_X_FORWARDED_PORT'];
+		}
+
+		$isInvalidCombination = $scheme === 'http' && $port === 443 || $scheme === 'https' && $port === 80;
```
Majkl578: Missing brackets.

Anyhow I think this should not be tested at all.

src/Http/RequestFactory.php
grongor: well in that case it will add one indention:
```
if ($usingTrustedProxy) {
	if (isset($_SERVER['HTTP_X_FORWARDED_FOR'])) {
		$remoteAddr = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_FOR'])));
	}

	if (isset($_SERVER['HTTP_X_FORWARDED_HOST'])) {
		$remoteHost = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_HOST'])));
	}

	if (!empty($_SERVER['HTTP_X_FORWARDED_PORT'])) {
		$port = (int) $_SERVER['HTTP_X_FORWARDED_PORT'];
	}
}
```

Or am I missing something?

src/Http/RequestFactory.php
grongor: Well, it work's, but I guess I should test the null value explicitly.

src/Http/RequestFactory.php
grongor: Good point, I'll change foreach to array_filter

src/Http/RequestFactory.php
grongor: I omitted them intentionally - it works well. Anyway if we don't test it there then we can endup with strange urls like `https://localhost:80`. Of course we can decide that this is ok and it's up to the developer to take care of these invalid links. The server would have to be badly missconfigured to produce such a strange combination of scheme and port.

src/Http/RequestFactory.php
```diff
@@ -60,25 +60,55 @@ public function setProxy($proxy)
...
+			if (isset($_SERVER['HTTP_X_FORWARDED_HOST'])) {
+				$remoteHost = trim(current(explode(',', $_SERVER['HTTP_X_FORWARDED_HOST'])));
 			}
 		}
 
-		// path & query
+		$isInvalidCombination = $scheme === 'http' && $port === 443 || $scheme === 'https' && $port === 80;
+		if ($port !== null && !$isInvalidCombination) {
+			$url->setPort($port);
+		}
```
JanTvrdik: How is that invalid? I always though that protocol is port independent.

JanTvrdik:
Should I point out all the issues or should I let @dg rewrite it?

src/Http/RequestFactory.php
grongor: That is true, however in that case it doesn't really makes sense. I think that a lot of browsers would choose to do that anyway. But if you think that this shouldn't be there I can remove it - I just thought that it would be for the better.

grongor:
Please point out the issues, otherwise I won't learn what's wrong with this contribution :)

src/Http/RequestFactory.php
```diff
@@ -60,25 +60,55 @@ public function setProxy($proxy)
 	 */
 	public function createHttpRequest()
 	{
+		$remoteAddr = !empty($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
+		$remoteHost = !empty($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
+
+		// use real client address and host if trusted proxy is used
+		$usingTrustedProxy = (bool)count(array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
```
JanTvrdik: There should be space after `(bool)`, the `count` call should be removed.

src/Http/RequestFactory.php
JanTvrdik: Yes I think it should be removed. Its too smart.

src/Http/RequestFactory.php
```diff
@@ -218,4 +231,38 @@ public function createHttpRequest()
 		return new Request($url, NULL, $post, $files, $cookies, $headers, $method, $remoteAddr, $remoteHost, $rawBodyCallback);
 	}
 
+	/**
+	 * @param bool $usingTrustedProxy
```
JanTvrdik: There should be two spaces between `@param` and `bool` (to align with `return`). The name of the parameter (`$usingTrustedProxy`) should be removed.

src/Http/RequestFactory.php
```diff
@@ -218,4 +231,38 @@ public function createHttpRequest()
...
+	 *
+	 * @return string 'http' or 'https'
+	 */
+	private function getScheme($usingTrustedProxy)
+	{
+		if ($usingTrustedProxy) {
+			if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
+				if (strcasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') === 0) {
```
JanTvrdik: Those two condition should be merged, possible rewritten as ternary expression.

src/Http/RequestFactory.php
```diff
@@ -218,4 +231,38 @@ public function createHttpRequest()
...
+
+			if (!empty($_SERVER['HTTP_X_FORWARDED_PORT'])) {
+				$port = (int) $_SERVER['HTTP_X_FORWARDED_PORT'];
+
+				if ($port === 443) {
+					return 'https';
+				} elseif ($port === 80) {
+					return 'http';
+				}
+			}
```
JanTvrdik: AFAIK you cannot derive schema from port. How does Symfony and Zend handle this?

src/Http/RequestFactory.php
```diff
@@ -218,4 +231,38 @@ public function createHttpRequest()
...
+		}
+
+		if (!empty($_SERVER['HTTPS']) && strcasecmp($_SERVER['HTTPS'], 'off') !== 0) {
+			return 'https';
+		}
+
+		return 'http';
```
JanTvrdik: Possibly rewrite to ternary expression.

src/Http/RequestFactory.php
grongor: ****, you are right, I again blindly copied the code from @Majkl578 :D

JanTvrdik:
Some more thoughts – I would try to remove the `getSchema` method entirely (inline in as it was before) and move the proxy-specific logic to where it was before.

src/Http/RequestFactory.php
grongor: Where did you get these information? I was following the [coding standards](https://nette.org/en/coding-standard). Omitting parameter name breaks PhpStorm and is not recommended by phpDocumentator - are you really sure that we want that?

JanTvrdik:
One of reasons to keep to the proxy logic together is https://github.com/nette/http/issues/20

src/Http/RequestFactory.php
grongor: I think it's the same case as with the conflicting scheme and ports. If we decide that we don't want the code to be that clever than we can remove this as well. It only handles badly configured proxies anyway.

src/Http/RequestFactory.php
```diff
@@ -218,4 +231,38 @@ public function createHttpRequest()
 		return new Request($url, NULL, $post, $files, $cookies, $headers, $method, $remoteAddr, $remoteHost, $rawBodyCallback);
 	}
 
+	/**
```
JanTvrdik: There should be two empty lines between two methods.

src/Http/RequestFactory.php
JanTvrdik: > Where did you get these information?

By looking at Nette code.


> I was following the coding standards.

Nope. Citation „The @param annotation is followed by two spaces, type and optionally by space and description.“

> Omitting parameter name breaks PhpStorm and is not recommended by phpDocumentator

This is well known but **consistency is a lot more important**. I even wrote Nette RFC to change that but there is unresolved issue with alignment.

grongor:
Thanks @JanTvrdik - I updated the code according to you suggestions and it's much easier :-) I also got rid of the "smart" scheme/port alignment. I hope I didn't miss something again ;-)

src/Http/RequestFactory.php
```diff
@@ -185,24 +187,33 @@ public function createHttpRequest()
 			}
 		}
 
+		$remoteAddr = !empty($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
+		$remoteHost = !empty($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
 
-		$remoteAddr = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
-		$remoteHost = isset($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
+		// use real client address and host if trusted proxy is used
+		$usingTrustedProxy = (bool)array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
```
hrach: do a space after the casting

grongor: ****, I tried so hard ... it's hard to notice these things when all other projects I'm working on use exactly opposite standards ...

hrach: still, you do an awesome job! thank you!

src/Http/RequestFactory.php
Majkl578: Sorry, count was a leftover, originally I had it like `count(...) === 0;` but then I changed my mind. :)

src/Http/RequestFactory.php
```diff
@@ -185,24 +187,33 @@ public function createHttpRequest()
...
 		}
 
+		$remoteAddr = !empty($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
+		$remoteHost = !empty($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
 
-		$remoteAddr = isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
-		$remoteHost = isset($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
+		// use real client address and host if trusted proxy is used
+		$usingTrustedProxy = (bool) array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
+			return $remoteAddr !== NULL && Helpers::ipMatch($remoteAddr, $proxy);
```
Majkl578: This `$remoteAddr !== NULL` is redundant, it should go outside the callback.

src/Http/RequestFactory.php
```diff
@@ -185,24 +187,34 @@ public function createHttpRequest()
 			}
 		}
 
+		$remoteAddr = !empty($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : NULL;
+		$remoteHost = !empty($_SERVER['REMOTE_HOST']) ? $_SERVER['REMOTE_HOST'] : NULL;
```
JanTvrdik: Why change `isset` to `!empty`? Isset worked OK for years, is faster to execute, can be in PHP 7 rewritten with `??` and does not consider `0` as empty.

src/Http/RequestFactory.php
```diff
@@ -76,9 +76,11 @@ public function createHttpRequest()
 			} elseif (isset($_SERVER['SERVER_PORT'])) {
 				$url->setPort((int) $_SERVER['SERVER_PORT']);
 			}
+		} else if (!empty($_SERVER['SERVER_PORT'])) {
+			$url->setPort((int) $_SERVER['SERVER_PORT']);
```
JanTvrdik: What's the point of having port without host? Does the code solve a real world issue? If not, it should be removed.

src/Http/RequestFactory.php
grongor: Because I don't think that having an empty host or address makes sense. Is that wrong assumption?

src/Http/RequestFactory.php
grongor: Yeah I guess not, you are right, thanks.

src/Http/RequestFactory.php
Majkl578: Having invalid value like '0' here makes no sense either and it's not checked anyway, maybe it could be.

grongor: That's true but I guess that we don't want to check webserver settings there. I just though that this is a little bit better (to ignore empty header rather the somehow try to use it).

dg:
@grongor thanks, merged 805ff25f95e491a618d250f2833a370f243c3032

PR #82, opened 2016-03-01
Do not suppress 'permission denied' warnings in FileUpload::move()

OndrejSlamecka:
This commit adds more comprehensible error reporting to the `move()` method in the following scenarios:

* Say the program has insufficient permissions to create the directory `dirname($dest)`. With this commit the `move()` method correctly raises `mkdir(): Permission denied` warning instead of warning raised by `move_uploaded_file`: `failed to open stream: No such file or directory` with previous code.
* Or the already existing file at `$dest` may not be deleted with the current permissions. With this commit a correct `Permission denied` warning is again raised by `unlink` instead of warning raised by `move_uploaded_file` with previous code.

### DIFF
--- a/src/Http/FileUpload.php
+++ b/src/Http/FileUpload.php
@@ -149,7 +149,7 @@ class FileUpload extends Nette\Object
 		@mkdir($dir, 0777, TRUE); // @ - dir may already exist
 		@unlink($dest); // @ - file may not exists
 		if (!is_dir($dir)) {
-			throw new \Nette\InvalidStateException("Permission denied: directory '$dir' cannot be created.");
+			throw new \Nette\InvalidStateException("Directory '$dir' cannot be created.");
 		}
 		if (file_exists($dest) && !is_writable($dest)) {
 			throw new \Nette\InvalidStateException("Permission denied: file '$dest' cannot be removed.");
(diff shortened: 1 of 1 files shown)


hrach:
You solution suffers with race condition problem.

enumag:
Then how about keeping the original code and adding `if (file_exists($dest)) { throw new ... }`?

OndrejSlamecka:
Good point and good suggestion, thank you. The commit above should prevent the problem at least to the level the previous code prevented such problems (but I wrote it with tablet, no guarantees :- )

hrach:
Are you sure, that only reason why directory has not been created is the permission problem? :)

OndrejSlamecka:
Right, inode limit comes to mind... I made the text more general with the commit above.

JanTvrdik:
See https://github.com/nette/tracy/issues/133

dg:
What about call mkdir only when dir doesn't exist and remove @?

Majkl578:
That wouldn't be atomic though.

dg:
This issue is about suppressing, isn't it?

Majkl578:
I replied to your previous comment:
> call mkdir only when dir doesn't exist

dg:
I understand.

PR #84, opened 2016-03-03
Use better detection for started session

TomasVotruba:
Constant `SID` doesn't detect active session properly and might return `TRUE` for inactive session.
Since PHP 5.4 there is better way to detect: [`session_status()`](http://php.net/manual/es/function.session-status.php)

```php
session_start();
var_dump(defined('SID')); // TRUE
var_dump(session_status() === PHP_SESSION_ACTIVE); // TRUE

session_destroy();
var_dump(defined('SID')); // TRUE
var_dump(session_status() === PHP_SESSION_ACTIVE); // FALSE
```

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -404,7 +404,7 @@ class Session extends Nette\Object
 				$cookie[substr($key, 7)] = $value;
 
 			} else {
-				if (defined('SID')) {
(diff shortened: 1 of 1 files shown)


dg:
Could you change commit message to something like `Session: …`?

TomasVotruba:
Sure, I've added the prefix

TomasVotruba:
Thank you

PR #85, opened 2016-03-07
Fix "Session object destruction failed"

luhla:
Close session before deleting data fix Session object destruction failed.


enumag:
It's this simple to fix that error? I feel ashamed.

JanTvrdik:
@dg this should be closed in favor of https://github.com/nette/http/pull/86

PR #86, opened 2016-03-07
Fix "Session object destruction failed" - update

luhla:


luhla:
Sorry, have to study how run tests before pull request...

hrach:
And try to fix the current one, not to open another :)

src/Http/Session.php
```diff
@@ -217,9 +217,9 @@ public function regenerateId()
 				throw new Nette\InvalidStateException('Cannot regenerate session ID after HTTP headers have been sent' . ($file ? " (output started at $file:$line)." : '.'));
 			}
 			if (session_id() !== '') {
+				session_write_close();
```
enumag: Maybe the `session_write_close();` should be before the if. That should fix the failing test as well.

luhla: I did try it before, but result was also failed checks. Solution is session_write_close(); before session_regenerate_id(TRUE); in my case but I wouldnt do annoing tests here.

dg:
Can you explain what is it good for?

luhla:
Getting "session_regenerate_id(): Session object destruction failed" when calling before session_write_close(); 
But i can`t call session_write_close(); before session_id().

dg:
Is it needed to call session_write_close() when session_id() === '' ?

luhla:
Getting error when using:
```
if (session_id() !== '') {
        session_regenerate_id(TRUE);
}
```
Everithing fine when using: 
```
if (session_status() === PHP_SESSION_ACTIVE ) {
        session_regenerate_id(TRUE);
}
```

dg:
IMHO session_write_close(); should be in condition too.

dg:
Did you change the order of session_write_close and session_regenerate_id intentionally?

luhla:
Make no sense for me, where should be session_write_close()  "in condition too" ? 
After session_regenerate_id(TRUE) it is unnecessary because there is another session_write_close() after condition. But it should be always before $backup = $_SESSION; 
Giving session_write_close() before session_regenerate_id(TRUE) is not obviously good idea.

dg:
> After session_regenerate_id(TRUE) it is unnecessary because there is another session_write_close() after condition

Why must be another after condition?

luhla:
Well, thats true, it might not, when seesion status is not active, nothing will be writen to session. I wil fix it.

dg:
Thanks!

luhla:
Thanks for nothing, you did the solution. :-)

Issue #87, opened 2016-03-09
RequestFactory: Possible remoteAddr spoofing


HonzaCZ:
Load balancers or proxies adds client IP to the end of X-Forwarded-For header.

So if site is behind trusted proxy, we set it by setProxy method and client sends spoofed X-Forwarded-For header then existing RequestFactory code interprets it as real client IP. Because proxy adds his real IP to the end but RequestFactory code gets the first IP from $_SERVER["HTTP_X_FORWARDED_FOR"] array.

Correct solution should be that we check $_SERVER["HTTP_X_FORWARDED_FOR"] array from the end compare to known trusted proxy array (set by setProxy) and use endmost IP that doesn't match any of know proxy IPs.

OK Example:
Site is behind 2 consecutive load balancers: 10.0.0.1 and 10.0.0.2. Clients real IP is 192.168.1.1.
From client there is no X-Forwarded-For header. First proxy set X-Forwarded-For to 192.168.1.1. Second proxy appends IP of first proxy - header will be X-Forwarded-For: 192.168.1.1, 10.0.0.1.
In this case, everything would be alright - we take first IP and it equals to real client IP, but...

Fake IP Example:
Situation as same as previous example but client sends spoofed X-Forwarded-For header. It sends e.g. 172.16.0.1 in that header.
First proxy appends his real IP, second proxy appends IP of first proxy. We have X-Forwarded-For: 172.16.0.1, 192.168.1.1, 10.0.0.1
So RequestFactory uses fake IP as reference.

milo:
Merged cae5d68c927c5502de8aa0c9992e4b968bedf49c

PR #88, opened 2016-03-09
RequestFactory: Fixed possible remoteAddr spoofing (issue #87)

HonzaCZ:
- bugfix
- issues - #87 
- documentation - not needed
- BC break - no

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -203,11 +203,20 @@ class RequestFactory extends Nette\Object
 			}
 
 			if (!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
-				$remoteAddr = trim(explode(',', $_SERVER['HTTP_X_FORWARDED_FOR'])[0]);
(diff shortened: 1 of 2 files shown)


src/Http/RequestFactory.php
```diff
@@ -203,11 +203,18 @@ public function createHttpRequest()
 			}
 
 			if (!empty($_SERVER['HTTP_X_FORWARDED_FOR'])) {
-				$remoteAddr = trim(explode(',', $_SERVER['HTTP_X_FORWARDED_FOR'])[0]);
+				$xForwardedForWithoutProxies = array_filter(explode(',', $_SERVER['HTTP_X_FORWARDED_FOR']), function ($ip) {
+					return !in_array(trim($ip), $this->proxies);
```
fprochazka: imho the the third argument of `in_array()` should be also used

HonzaCZ: I have better solution now. Working on tests. Commit ASAP.

PR #92, opened 2016-05-01
Helper for parsing headers attributes

JanTvrdik:
I'm missing a helper for parsing headers, i.e something that turns `attachment; filename="example.php"` into `[filename => 'example.php']`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
 	],
 	"require": {
 		"php": ">=5.6.0",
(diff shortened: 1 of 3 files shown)


JanTvrdik:
The stuff is so fucked up that even browsers can't implement it correctly http://greenbytes.de/tech/tc/httpcontenttype/

f3l1x:
I like an idea. Any progress here?

Issue #91, opened 2016-04-29
RequestFactory, REQUEST_URI and absolute URLs


ondrej-tuhacek:
Recently I installed this addon to Google Chrome:

[https://chrome.google.com/webstore/detail/host-switch-plus/bopepoejgapmihklfepohbilpkcdoaeo](https://chrome.google.com/webstore/detail/host-switch-plus/bopepoejgapmihklfepohbilpkcdoaeo)

its purpose is something like **etc/hosts** file with one single exception. If used, it sends HTTP requests like:

`GET http://www.example.com/ HTTP/1.1`

instead of 

`GET / HTTP/1.1`

which seems to be valid according to [https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html](https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html)

But Nette shows "no routes" and gives me 404.

Because RequestFactory.php where $_SEVER['REQUEST_URI'] is parsed cannot handle absolute URLs. I quick-fixed this simply by modifying urlFilters from:

```
	/** @var array */
	public $urlFilters = array(
		'path' => array('#/{2,}#' => '/'), // '%20' => ''
		'url' => array(), // '#[.,)]\z#' => ''
	);
```

to:

```
	/** @var array */
	public $urlFilters = array(
		'path' => array('#/{2,}#' => '/'), // '%20' => ''
		'url' => array('#^(http|ftp)s?://[^/]+/#' => '/'), // '#[.,)]\z#' => ''
	);
```

and it works, but I'm not sure if this is the best way to "fix" this.

Majkl578:
According to [RFC 7230 ch. 5.3Request Target](https://tools.ietf.org/html/rfc7230#section-5.3), this is not valid.
> When making a request directly to an origin server, other than a
> CONNECT or server-wide OPTIONS request (as detailed below), a client
> MUST send only the absolute path and query components of the target
> URI as the request-target.

ondrej-tuhacek:
Yes, you are right. In the most common used scenario it should be the **absolute path**, but in section 5.3.2 there is also **absolute form = absolute-URI** for proxies (and maybe the chrome add-on acts as a proxy - but I'm still not sure). And furthermore there is this sentence:

> To allow for transition to the absolute-form for all requests in some
   future version of HTTP, a server **MUST accept the absolute-form in
   requests, even though HTTP/1.1 clients will only send them in
   requests to proxies**.

[...]

PR #93, opened 2016-05-06
remove absolute URI from $_SERVER['REQUEST_URI'] (issue #91)

ondrej-tuhacek:


src/Http/RequestFactory.php
```diff
@@ -81,7 +81,7 @@ public function createHttpRequest()
 		}
 
 		// path & query
-		$requestUrl = isset($_SERVER['REQUEST_URI']) ? $_SERVER['REQUEST_URI'] : '/';
+		$requestUrl = isset($_SERVER['REQUEST_URI']) ? Strings::replace($_SERVER['REQUEST_URI'], '#^\w+://[^/]+#i') : '/';
```
JanTvrdik: I think it's better to make it more readable by splitting it into lines.

~~~php
$requestUrl = isset($_SERVER['REQUEST_URI']) ? $_SERVER['REQUEST_URI'] : '/';
$requestUrl = Strings::replace($requestUrl, '#^\w+://[^/]+#i');
~~~

The pattern could be optimized by removing pointless case-insensitive flag and avoiding any possible backtracking

~~~php
$requestUrl = isset($_SERVER['REQUEST_URI']) ? $_SERVER['REQUEST_URI'] : '/';
$requestUrl = Strings::replace($requestUrl, '#^\w++://[^/]++#');
~~~

With removed backtracking, we may probably replace `Strings::replace` with faster `preg_replace`

~~~php
$requestUrl = isset($_SERVER['REQUEST_URI']) ? $_SERVER['REQUEST_URI'] : '/';
$requestUrl = preg_replace('#^\w++://[^/]++#', '', $requestUrl);
~~~

ondrej-tuhacek: OK, you're right. I have modified it. Thanks.

src/Http/RequestFactory.php
```diff
@@ -82,6 +82,7 @@ public function createHttpRequest()
 
 		// path & query
 		$requestUrl = isset($_SERVER['REQUEST_URI']) ? $_SERVER['REQUEST_URI'] : '/';
+		$requestUrl = preg_replace('#^\w++://[^/]++#', '', $requestUrl);
```
Majkl578: Is this actually RFC compliant? What if I send mismatching Host header or use different protocol than used in here? Shouldn't we strip it only if it is a match and throw exception otherwise?

Majkl578: Example:
```
$ telnet google.com 80
Trying 216.58.209.174...
Connected to google.com.
Escape character is '^]'.
GET https://google.com/ HTTP/1.1
Host: google.cz

HTTP/1.1 400 Bad Request
Content-Type: text/html; charset=UTF-8
Content-Length: 1555
Date: Fri, 06 May 2016 13:57:07 GMT
```

JanTvrdik: See [how Symfony handles this](https://github.com/symfony/http-foundation/blob/fac93f4c3c74c3e4609f98c2b98918dc6f1c4c3c/Request.php#L1719-L1725). They replace it only if it matches current schema and host. But I haven't seen any specification which would define how this should actually be handled.

PR #94, opened 2016-05-21
Implemented RFC 7239 - "Forwarded HTTP Extension"

patrickkusebauch:
- feature
- issues - #20 
- documentation - not needed - just see RFC 7239
- BC break - no

RequestFactory can now take "RemoteAddress", "RemoteHost", "port" and "scheme" from the proxy "FORWARDED" header.


src/Http/RequestFactory.php
```diff
@@ -194,30 +194,54 @@ public function createHttpRequest()
...
+				foreach ($forwardParams as $forwardParam) {
+					$param = explode("=", $forwardParam);
+					$proxyParams[trim($param[0])][] = trim($param[1]);   //e.g. array['for'][0] = 192.168.0.1
+				}
 
-			if (!empty($_SERVER['HTTP_X_FORWARDED_PORT'])) {
-				$url->setPort((int) $_SERVER['HTTP_X_FORWARDED_PORT']);
-			}
+				if (isset($proxyParams['for'])) {
+					$remoteAddr = explode(':', $proxyParams['for'][0])[0];
```
Majkl578: This will not work with IPv6.

src/Http/RequestFactory.php
```diff
@@ -194,30 +194,54 @@ public function createHttpRequest()
...
-
 		if ($usingTrustedProxy) {
-			if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
-				$url->setScheme(strcasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') === 0 ? 'https' : 'http');
-			}
+			if(!empty($_SERVER['HTTP_FORWARDED'])) {
+				$forwardParams = preg_split('/[,]|[;]/', $_SERVER['HTTP_FORWARDED']);
+				foreach ($forwardParams as $forwardParam) {
+					$param = explode("=", $forwardParam);
+					$proxyParams[trim($param[0])][] = trim($param[1]);   //e.g. array['for'][0] = 192.168.0.1
```
Majkl578: Parameter names must be treated in case-insensitive manner.

src/Http/RequestFactory.php
```diff
@@ -194,30 +194,54 @@ public function createHttpRequest()
 		$usingTrustedProxy = $remoteAddr && array_filter($this->proxies, function ($proxy) use ($remoteAddr) {
 			return Helpers::ipMatch($remoteAddr, $proxy);
 		});
-
 		if ($usingTrustedProxy) {
-			if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
-				$url->setScheme(strcasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') === 0 ? 'https' : 'http');
-			}
+			if(!empty($_SERVER['HTTP_FORWARDED'])) {
+				$forwardParams = preg_split('/[,]|[;]/', $_SERVER['HTTP_FORWARDED']);
```
Majkl578: Semicolin is used as a forward-pair delimiter whereas comma is used as a field separator. I'm afraid current implementation won't be able to distinguish it...

src/Http/RequestFactory.php
Majkl578: Also this doesn't properly handle quoted values.

src/Http/RequestFactory.php
patrickkusebauch: It is not. However it should not be a problem, as I believe there is no need for the distinction.
There two headers would produce the same result array:
for=192.0.2.43, for=198.51.100.17
for=192.0.2.43; for=198.51.100.17

I just need to split the "key-value" pairs and put them in the proper sub-array.

src/Http/RequestFactory.php
patrickkusebauch: Parameter names must be treated in case-insensitive manner.
solved by commit 44660bd

patrickkusebauch: Quoted values in commit 11461e6

src/Http/RequestFactory.php
patrickkusebauch: IPv6 support in commit 11461e6

JanTvrdik:
possible related https://github.com/nette/http/pull/92

src/Http/RequestFactory.php
JanTvrdik: This can be simplified to `preg_split('#[,;]#')`

src/Http/RequestFactory.php
```diff
@@ -194,30 +194,64 @@ public function createHttpRequest()
...
+				foreach ($forwardParams as $forwardParam) {
+					$param = explode("=", $forwardParam);
+					$proxyParams[strtolower(trim($param[0]))][] = trim($param[1], "\"\t\n\r\0\x0B");   //e.g. array['for'][0] = 192.168.0.1
+				}
```
JanTvrdik: This can be simplified to 

~~~php
foreach ($forwardParams as $forwardParam) {
    list($key, $value) = explode('=', $forwardParam, 2) + [1 => NULL];
    $proxyParams[strtolower(trim($key))] = trim($value, " \t\"");
}
~~~

patrickkusebauch: Integrated in commit a2c7a79

src/Http/RequestFactory.php
patrickkusebauch: Integrated in commit a2c7a79

patrickkusebauch:
I know that it keeps failing on php5.6 with composer update --no-interaction --prefer-dist --prefer-lowest --prefer-stable. However this cannot be redeemed as the tests have to change to keep up with the versions. Furthermore they are test unrelated to this pull request.

patrickkusebauch:
Any other problems to be fixed with this PR? 
Missing tests?
Or can I safely squash it to be merged?

dg:
Is it ready for merge?

patrickkusebauch:
Since there were no comments to the PR for more than 3 weeks, I would assume nobody has any more problems with it.
So yes, it is ready for merge.

dg:
So thanks!

patrickkusebauch:
I thank you for letting me be a part of the development of Nette. :)

PR #96, opened 2016-07-27
better URI syntax complience

t0his:
Nette\Http\Url::getHostUrl() returns '//' when no scheme or host is suplied.
Which is slightly annoing for me and it goes against URI syntax 'scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]'.

### DIFF
--- a/src/Http/Url.php
+++ b/src/Http/Url.php
@@ -376,7 +376,7 @@ class Url implements \JsonSerializable
 	 */
 	public function getHostUrl()
 	{
-		return ($this->scheme ? $this->scheme . ':' : '') . '//' . $this->getAuthority();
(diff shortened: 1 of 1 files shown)


Majkl578:
Leading `//` is mandatory if authority is present in the URI (regardless if scheme is present or not), as per [RFC 3986](https://tools.ietf.org/html/rfc3986#section-4.2).

Also PSR-7 [follows this](http://www.php-fig.org/psr/psr-7/#3-5-psr-http-message-uriinterface):
> If an authority is present, it MUST be prefixed by "//".

dg:
For `http:` it should be omitted, for `file:` is should not https://en.wikipedia.org/wiki/File_URI_scheme.

PR #102, opened 2016-09-15
Session: nette-browser cookie may be sent with secure flag

PavelJurasek:


src/Http/Session.php
```diff
@@ -545,7 +545,7 @@ private function sendCookie()
 		);
 		$this->response->setCookie(
 			'nette-browser', $_SESSION['__NF']['B'],
-			Response::BROWSER, $cookie['path'], $cookie['domain']
+			Response::BROWSER, $cookie['path'], $cookie['domain'], $cookie['secure']
```
JanTvrdik: * Why not pass httponly option as well?
* This has been removed in https://github.com/nette/http/commit/0aed1a2bb65f42cb8a32f3909fd593538ac52f57, not sure why

PavelJurasek: You are right, I'll add httponly too.

dg:
What's the reason?

PavelJurasek:
Mixed content is being sent when site in on HTTPS.

dg:
It is error message? In which browser?

PavelJurasek:
There is no error nor warning in browsers. At least not when HSTS is used. 
Mozilla Observatory: 
screen

dg:
Closed by #103

PR #103, opened 2016-09-21
Session: removed cookie nette-browser

dg:
Session: removed cookie nette-browser and support for expiration 'until the browser is closed'

It is not supported by browsers.


JanTvrdik:
I've never used it personally.

PR #104, opened 2016-09-23
Session: used static access to $started

CZechBoY:
Deprecated error was triggered.
$this->started instead of self::$started was used.


dg:
Thanks

PR #107, opened 2016-10-03
IResponse: added HTTP response codes from accepted RFC 6585

matyx:
RFC: https://tools.ietf.org/html/rfc6585


dg:
Do you think this is useful?

matyx:
@fprochazka Yes, I will add them.
@dg I need only 429 - Too many requests (for API). I added all from the RFC only because of completness

fprochazka:
@dg it's better to have those codes as constants. And it's weird to have half defined in nette and half in custom class.

Majkl578:
There is even more codes missing here, see Symfony's Response constants: https://github.com/symfony/symfony/blob/c0a26bc/src/Symfony/Component/HttpFoundation/Response.php#L21-L82

matyx:
I've added all missing response codes

rostenkowski:
:tophat:

rostenkowski:
As for community-driven development this a good pull request!

:+1:

fprochazka:
@matyx awesome :+1:

PR #108, opened 2016-10-14
Make sure __NF in session is array

juniwalk:
There is a problem that when request fails in specific moment, sessions gets destroyed and the `$nf` variable is then equal to `0`. This allows the session to recover without throwing error.


dg:
Thanks

PR #116, opened 2017-01-18
appveyor: upgrade php to version 7.1

foxycode:
- bug fix? yes
- new feature? no
- BC break? no

Also changed x86 to x64 version. If it's ok, I make PR for all other packages.


dg:
Thanks

dg:
There is missing `/archives/` #117

foxycode:
Oh, I see, thanks.

PR #126, opened 2017-07-31
Session: initialize $started property in constructor

JanTvrdik:
- bug fix? yes
- new feature? depends
- BC break? no, hopefully

This fixes the issue when new instance of DI container (with session service) is created after session has been manually aborted via `session_abort()` due to database transaction failure.

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -26,7 +26,7 @@ class Session
 	private $regenerated = false;
 
 	/** @var bool  has been session started? */
-	private static $started;
+	private static $started = false;
 
 	/** @var array default configuration */
 	private $options = [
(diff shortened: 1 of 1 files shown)


dg:
Is possible to change session name between two started and aborted sessions?

JanTvrdik:
@dg I'm not sure I understand the point of your question. But it seems reasonable to assume that both session name and session id could be changed.

JanTvrdik:
@dg I'm not proposing to support multiple `Sessions` instances existing at the same time. I only need to support multiple instances of `Sessions` where the second one exists after the first one has been garbage collected & sessions has been aborted (which should rollback all changes).

dg:
Ok I understand.

PR #131, opened 2017-10-20
HttpExtension: added ability to specify list of proxies

CZechBoY:
- bug fix? no
- new feature? yes
- BC break? no
- doc PR: maybe

Hi,
if i want to specify multiple proxies then it is not possible in current version.
This PR adds ability to specify multiple proxies for one application.


dg:
Tests are failing.

CZechBoY:
ok, now i see that both variants are already supported and tested too.

PR #132, opened 2017-11-24
FileUpload: Add Webp to isImage

holantomas:
- bug fix? no
- new feature? yes
- BC break? no

According to content types in `Nette\Utils\Image` added `image/webp` to `isImage()` method.


dg:
Thanks

Issue #129, opened 2017-09-01
Restarting session after ID regeneration looses session


KacerCZ:
- bug report? yes
- feature request? no
- version: 2.4.6

### Description
I found problem with restarting session after session ID was regenerated.
This is because `\Nette\Http\Session::start()` takes session ID always from request cookie. Invocation of `\Nette\Http\Session::regenerateId()` changes session ID. Closing session and starting session again changes session ID to value which is stored in cookie, but this session ID is already deleted.

### Steps To Reproduce
```php
$session = new Session();
$session->start(); // Session ID is taken from cookie.
$session->regenerateId(); // Session ID is changed.
$session->close(); // Session with old ID is deleted, session with new ID is written.
$session->start(); // Session ID is taken from cookie - PROBLEM 
// it is old session ID, not the regenerated one and all session data are lost
```

dg:
For information ID is taken from cookie since https://github.com/nette/nette/issues/965.

Can you send PR?

KacerCZ:
What is correct fix for this issue? Always get session ID using `session_id()` in methods `start()` and `exists()`?

PR #133, opened 2018-03-06
Take session ID from session_id() instead of cookie.

KacerCZ:
- bug fix? yes
- new feature? no
- BC break? no

Fixes https://github.com/nette/http/issues/129
Session ID is taken from `session_id()` instead of cookie.

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -77,7 +77,8 @@ class Session
 
 		$id = session_id();
 		if (!is_string($id) || !preg_match('#^[0-9a-zA-Z,-]{22,256}\z#i', $id)) {
-			session_id('');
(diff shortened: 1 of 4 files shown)


KacerCZ:
AppVeyor fail is not related to this PR.

src/Http/Session.php
```diff
@@ -75,11 +75,10 @@ public function start()
 
 		$this->configure($this->options);
 
-		$id = $this->request->getCookie(session_name());
-		if (is_string($id) && preg_match('#^[0-9a-zA-Z,-]{22,256}\z#i', $id)) {
-			session_id($id);
-		} else {
-			unset($_COOKIE[session_name()]);
+		$id = session_id();
```
dg: I tested it and it returns empty string when session is not started.

PR #134, opened 2018-03-06
Cannot set session cookie parameters for already started session

mabar:
- bug fix? yes
- new feature? no
- BC break? probably not

Fixes `Warning session_set_cookie_params(): Cannot change session cookie parameters when session is active`
It happens when Session::setOptions() is called multiple times.
Can be tested with php 7.2+ only. In previous versions it failed silently (see https://bugs.php.net/bug.php?id=75650&thanks=2)


mabar:
@dg Thanks but nothing changed :(
It makes no sense to call `session_name($name)` or `session_set_cookie_params()` twice during one request even if parameters was changed.

dg:
Why not?

PR #135, opened 2018-03-11
HttpExtension: allow setup CSP in report only mode

PavelJurasek:
- bug fix? no
- new feature? yes
- BC break? no
- doc PR: will do if accepted

CSP can now be enabled in report only mode. PR to nette/application will be sent in just a minute.

### DIFF
--- a/src/Bridges/HttpDI/HttpExtension.php
+++ b/src/Bridges/HttpDI/HttpExtension.php
@@ -95,6 +95,13 @@ class HttpExtension extends Nette\DI\CompilerExtension
 		}
 
 		if (!empty($config['csp'])) {
+			$reportOnly = false;
+
+			if (array_key_exists('reportOnly', $config['csp'])) {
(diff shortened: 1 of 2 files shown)


dg:
The question is whether it makes sense to use the headers Content-Security-Policy and Content-Security-Policy-ReportOnly (with different settings) together?

PavelJurasek:
This commit allows only one of them to be enabled at a time. But I can actually see a case when I have a CSP policy and I want to experiment with more strict policy via Report only at the same time.

dg:
It would be better to add something like `csp-report` or `csp-report-only` section.

PavelJurasek:
In order to support both of them at the same time? I don't need this functionality at the moment, so maybe opening an issue with this feature description just to keep track of it would be enough for now?

dg:
It is good to think ahead. Will be possible to implement support for both header in a way that will co-exist with this solution?

PR #136, opened 2018-03-11
HttpExtension: allow setup CSP in restrictive and report mode at the …

PavelJurasek:
…same time

Solution for both modes of CSP.
Related to https://github.com/nette/http/pull/135.

### DIFF
--- a/src/Bridges/HttpDI/HttpExtension.php
+++ b/src/Bridges/HttpDI/HttpExtension.php
@@ -124,7 +124,8 @@ class HttpExtension extends Nette\DI\CompilerExtension
 		}
 	}
 
-	private function mapContentSecurityPolicies(array $csp): string
+
+	private function mapContentSecurityPolicies(array $csp)
 	{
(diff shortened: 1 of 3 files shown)


PavelJurasek:
@spaze Do you think it's OK to use the same nonce for CSP and CSP-ReportOnly headers?

dg:
Thanks!

PavelJurasek:
I thank you! 🙏 
I'll be happy to see new release as it has been over 6 months since last release.

dg:
I picked it in 2.4 too

spaze:
@PavelJurasek 
>Do you think it's OK to use the same nonce for CSP and CSP-ReportOnly headers?

It is. Also, nice addition, although `csp-report` might be a bit misleading, I'd suggest `csp-report-only` or something but it's probably too late.

dg:
I have changed option to `cspReportOnly` to be consistent with the rest of configuration.

PR #140, opened 2018-07-09
RequestFactory: SERVER_NAME has higher priority over HTTP_HOST

CZechBoY:
- bug fix 
- BC break? yes

Hi,
it is better to identify server host name by secure directive SERVER_NAME, because HTTP_HOST is not secure and can be changed by user.
https://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -70,7 +70,7 @@ class RequestFactory
 
 		// host & port
 		if (
-			(isset($_SERVER[$tmp = 'HTTP_HOST']) || isset($_SERVER[$tmp = 'SERVER_NAME']))
+			(isset($_SERVER[$tmp = 'SERVER_NAME']) || isset($_SERVER[$tmp = 'HTTP_HOST']))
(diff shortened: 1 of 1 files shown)


dg:
Does not cause it a BC break?

CZechBoY:
Yes, someone can depend on it/do not configure webserver proberly. I mention that BC break in PR header.

JanTvrdik:
I'm afraid this will break a lot of stuff.

JanTvrdik:
Related http://symfony.com/blog/security-releases-symfony-2-0-24-2-1-12-2-2-5-and-2-3-3-released#cve-2013-4752-request-gethost-poisoning

jakubboucek:
Yep, it can result to big BC break on webhostings with [`UseCanonicalName On`](https://httpd.apache.org/docs/2.4/mod/core.html#usecanonicalname) set.

Currently is default value `off`, but Apache 2.0 and older [was default value `On`](https://httpd.apache.org/docs/2.0/mod/core.html#usecanonicalname).

dg:
Closing due to big BC break,.

PR #144, opened 2018-08-13
HttpExtension: X-Frame-Options header

harmim:
- new feature
- BC break? no

Avoid processing `frames` config if header `X-Frame-Options` is already set in configuration. If `X-Frame-Options` header is already set, there is unnecessary invocation of `preg_match`, etc. This `frames` key option should not be processed at all in this case.


dg:
Thanks

dg:
But it seems to me unnecessary to save one preg_match when container is compiled…

PR #146, opened 2018-11-23
SessionExtension: added option handler to pass own SessionHandlerInterface

milo:
- new feature
- BC break? no

Own session handler cannot be set via `session.save_handler`, the `ini_set()` accepts string type only.
Can be solved by decorator extension:
```neon
decorator:
    Nette\Http\Session:
        setup:
            - setHandler(...)
```
but propesed way seems more suitable to me.


milo:
Thank you!

Issue #147, opened 2018-12-02
Request API vs IRequest interface


xificurk:
`Request` implements a couple of extra methods that are not defined in `IRequest` interface. Nette 3 deprecated autowiring of `Request` type, only autowiring via interface is now supported. This means that those extra methods are basically "inaccessible" for any consumer. I think those methods should be either added to the interface, or deprecated and eventually removed.

- `getReferer(): ?Url` - the interface already mentions this method in `@method` annotation, the annotation should be transformed to a proper method definition.
- `isSameSite(): bool` - this should be definitely added to the interface.
- `detectLanguage(array $langs): ?string` - this can be easily extracted to a separate service and removed from `Request`.

What are your thoughts? Shall I send a PR(s)?

Related to #137, #90

xificurk:
@dg Any thoughts? I think this needs to be resolved before stable 3.0 version.

dg:
For the next version there are plans to move to middleware and there will be probably a lot of changes. So currently I do not want make a BC break and we'll solve it the next time.

xificurk:
I really think this needs to be solved now because of BC break introduced in #90.

Please, tell me if I'm missing some crucial point. The following code works just fine in Nette 2.4.x:
```php
class Foo
{
    /** @var Nette\Http\Request */
    private $request;
 
   public function __construct(Nette\Http\Request $request)
    {
        $this->request = $request;
    }
 
    public function bar(): void
    {
        if ($this->request->isSameSite()) {
            // whatava
        }
    }
```
As far as I know, there is no way how to make it work in Nette 3.x. I can't use directly `Request` type and the `IRequest` interface does not declare `isSameSite()` method (and bunch of others).

dg:
I reverted fix for #90.

Issue #90, opened 2016-04-26
DI: Nette\Http\Request subclass is required instead of interface


iinfo-dev-mk:
When we want to you own service for http request, subclass of Nette\Http\Request is required

``` php
// contex is instance of Nette\DI\Container
$context->removeService("http.request");
// OwnRequest implements Nette\Http\IRequest
$context->addService("http.request", new OwnRequest());
```

it throws exception

Nette\InvalidArgumentException: Service 'http.request' must be instance of Nette\Http\Request, OwnRequest given

I think same problem is with all services from Nette\Http package. Only interface can be requested, not class.

Maybe problem is with DI container itself, when _removeService_ is called, I wan't to remove all information about particular service, it seems that in DI container still remains some information about removed service, what isn't desirable.

fprochazka:
There should be no references to `Nette\Http\Request` in Nette, since we've successfully replaced it with impl of it's interface and the application worked. You should first check your application, to make sure you're not requiring the class itself, instead of interface and then check the extensions you're using. Also maybe upgrade Nette, since the last referenced were removed few versions ago.

And also, the request object is a value object, that should in fact be a service. If you wanna work with multiple requests, you should use [Kdyby/RequestStack](https://github.com/Kdyby/RequestStack).

iinfo-dev-mk:
@fprochazka I'm using [this default extension](https://github.com/nette/http/blob/v2.3/src/Bridges/HttpDI/HttpExtension.php). After container was created, I tried to replace http request services with code above, and it failed with the exception. 
It worked fine with nette 2.2, with nette 2.3 it failed.

dg:
I remember that there was a break BC, when I changed it to interfaces, but I do not remember the details.

matej21:
@dg yes, there would be a BC break if someone requires Http\Request directly and not the interface.

dg:
Exactly.

iinfo-dev-mk:
Require directly Http\Request directly isn't good practice if DIC offers service name by default. But replace particular service with another one is valid requirement for the DIC. 

@dg No, see the issue description, there is no issue with the configuration, I need to use removeService, addService calls. Replacing services this way doesn't work correctly. Hence I also noticed if problem isn't in DI container itself, see my description.

dg:
I mean that you can change class in config.neon

```
services:
    http.request:
         class: Nette\Http\IRequest
```

and then replace it in bootstrap

iinfo-dev-mk:
[...]

PR #148, opened 2018-12-03
Response, IResponse: unified API (BC break)

xificurk:
- bug fix / new feature? a bit of both
- BC break? yes
- doc PR: don't think it's needed

`Response` API at the moment slightly differs from `IResponse`. This PR unifies both APIs. Also, explicitly marking all nullable parameters as nullable in parameter type hint.

### DIFF
--- a/src/Http/IResponse.php
+++ b/src/Http/IResponse.php
@@ -88,7 +88,7 @@ interface IResponse
 	 * Sets HTTP response code.
 	 * @return static
 	 */
-	function setCode(int $code);
+	function setCode(int $code, ?string $reason = null);
 
 	/**
 	 * Returns HTTP response code.
@@ -99,7 +99,7 @@ interface IResponse
 	 * Sends a HTTP header and replaces a previous one.
 	 * @return static
 	 */
-	function setHeader(string $name, string $value);
+	function setHeader(string $name, ?string $value);
 
 	/**
 	 * Adds HTTP header.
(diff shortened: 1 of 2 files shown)


dg:
`?type $var = null` is inconsistent with the rest of framework, where is only `type $var = null`.

xificurk:
I always assumed this was a hack for declaring nullable parameters from pre 7.1 times.

dg:
For PHP both declarations are identical. We can agree to change `type $var = null` => `?type $var = null` in the whole framework, but I think it is unnecessary.

xificurk:
OK, but #398dc22 is not enough, there is still a major incompatibility:
`Response::setHeader(string $name, ?string $value)` vs `IResponse::setHeader(string $name, string $value)`, plus `IResponse` is completely missing `deleteHeader()`.
And since it's currently not possible to require `Response` by specific type, it means there is no way to delete remove header.
That was actually the original motivation for this PR. I've got failing builds, at first because I was relying directly on `Response`, which was deprecated in #90. After replacing the specific implementation by interface, the build started to fail because static analysis (correctly) flagged calls `IResponse:.setHeader('Foo', null)` or `IResponse:.deleteHeader('Foo')` as an error.

dg:
I understand where problem is and #147 is opened.

There will certainly be much more similar problems in Nette and I'll try to fix them in the next version.

xificurk:
#147 revolves around `IRequest` vs `Request` API. This PR was about `IResponse` vs `Response` API, but the root cause is the same - BC break from #90.

PR #149, opened 2019-01-06
HTTPS and mod_remoteip

kncsvk:
- bug fix
- BC break? no

I have web server setup with HTTP proxy and Apache web server with module mod_remoteip (This mod replace REMOTE_ADDR to real client IP instead of proxy IP). 
HTTP works as expected but with HTTPS I have these problems:

With router:
`$router[] = new Route('<presenter>/<action>[/<id>]', 'Dashboard:default');`
Result URL was `https://example.com:80/`

With router:
`$router[] = new Route('https://%host%/<presenter>/<action>[/<id>]', 'Dashboard:default');`
Result in browser was redirect loop (`ERR_TOO_MANY_REDIRECTS`)

I searched in code and found that HTTP_X_FORWARDED* headers are ignored if they din't come from trusted proxy. After bypassing this check for HTTP_X_FORWARDED_PROTO and HTTP_X_FORWARDED_PORT router works again as expected.
I don't see any security problem if these two variables didn't come from trusted proxy.

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -265,8 +265,17 @@ class RequestFactory
 					}
 				}
 			}
-		}
+		} else {
+			if (!empty($_SERVER['HTTP_X_FORWARDED_PROTO'])) {
+				$url->setScheme(strcasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') === 0 ? 'https' : 'http');
+				$url->setPort($url->getScheme() === 'https' ? 443 : 80);
+			}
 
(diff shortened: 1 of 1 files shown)


mabar:
You can spam application with shitload of errors if app breaks with an incompatible url. In worst case would be e.g. possible send user an email with url of attacker website.

JanTvrdik:
You can always mark all proxies as trusted with `setProxy('0.0.0.0/0')` if your network setup ensures that your app is always behind trusted proxy.

dg:
@kncsvk does this work for you: (https://doc.nette.org/cs/3.0/configuring#toc-http-proxy)

```neon
http:
    proxy: 0.0.0.0/0
```

PR #153, opened 2019-03-17
Configure headers in response setup

mabar:
- BC break: no
- doc PR: not needed

Moved headers configuration from initialize() method of DIC to Response service setup.
Useful for applications with multiple http layer implementations (we use psr-7 in Apitte). It should prevent mixing headers from nette/http if not used for current request

### DIFF
--- a/src/Bridges/HttpDI/HttpExtension.php
+++ b/src/Bridges/HttpDI/HttpExtension.php
@@ -10,7 +10,6 @@ declare(strict_types=1);
 namespace Nette\Bridges\HttpDI;
 
 use Nette;
-use Nette\PhpGenerator\Helpers;
 use Nette\Schema\Expect;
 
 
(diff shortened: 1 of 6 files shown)


mabar:
I am not sure how to fix tests. I tested it in my app and it should work properly.

mabar:
@dg Ready

dg:
I'm afraid this is a BC break, which is not suitable for the patch version.

mabar:
I thinked about it too. It should be BC break only in case that someone depends on fact that headers are sent and Response is not instantiated by user's app.

nette\application users are ok, only users of standalone nette/http should be affected and just in case they send response incorrectly without Response class.

As far as I know, Apitte is the only http-related nette extension, which is (almost) completely separated from nette/http and it will be not affected by this change until next version.

Do you know an other case which could be problematic? This PR is non-blocking for me, so it could wait. Just an architectural problem.

mabar:
Maybe a compatibility-mode, which would be enabled by default? It would just initialize service in initialize() method, if enabled.

Something like [this](https://github.com/contributte/monolog/blob/v0.3.0/src/DI/MonologExtension.php#L136)

dg:
In case it is non-blocking for you, I will postpone it to 3.1.

mabar:
ok

mabar:
@dg Would this change be acceptable for v4.0? Response will not sent headers directly anyway and it would be helpful if it didn't interfere with other http libraries.

PR #156, opened 2019-06-18
Session: read_and_close configuration

### LINKED ISSUE #155: session - read_and_close
`session_start()` supports option `read_and_close` which could massively increase performance in apps with multiple simultaneous requests. It would be nice to allow configuration of this directive. I can send PR if needed.

https://www.php.net/manual/en/function.session-start.php

mabar:
Closes #155 

Should be the only option which is configurable only in session_start.


dg:
Because session is configured via setOptions(), it would seem better to me to use `setOptions('['read_and_close' => true])` than use constructor. What do you think?

mabar:
Okay. I will add check that read_and_close is only configured before session started

mabar:
@dg Ready.

dg:
Great, thanks!

PR #161, opened 2019-06-28
RequestFactory: set user and password to Url

### LINKED ISSUE #160: Request url does not contain user and password
Version: 3.0.1

### Bug Description
In nette 2.4 was set user and password to url [here](https://github.com/nette/http/blob/v2.4/src/Http/RequestFactory.php#L68-L69)

In 3.0, this code dissappeared so user and password are now always empty string.

### Expected Behavior
User and password should be set from `$_SERVER['PHP_AUTH_USER']` and `$_SERVER['PHP_AUTH_USER']` to `Url`

### Possible Solution
Add [here](https://github.com/nette/http/blob/master/src/Http/RequestFactory.php#L65) method setUserAndPassword

mabar:
Closes #160

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -65,6 +65,7 @@ class RequestFactory
 		$url = new Url;
 		$this->getServer($url);
 		$this->getPathAndQuery($url);
+		$this->getUserAndPassword($url);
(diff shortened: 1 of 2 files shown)


dg:
Thanks

PR #164, opened 2019-07-09
Rename general array type to mixed[].

janbarasek:
- new feature
- BC break? no

Simple rename general type to `mixed[]` for better code-style hinting in PhpStorm.

### DIFF
--- a/src/Http/UrlImmutable.php
+++ b/src/Http/UrlImmutable.php
@@ -36,7 +36,7 @@ use Nette;
  * @property-read string $absoluteUrl
  * @property-read string $authority
  * @property-read string $hostUrl
- * @property-read array $queryParameters
+ * @property-read mixed[] $queryParameters
  */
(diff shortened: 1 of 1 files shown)


dg:
> better code-style hinting

Can you explain what you mean?

janbarasek:
@dg In this case PhpStorm can hint methods automatically in loop by `@return` annotation:

Snímek obrazovky 2019-07-09 v 14 48 50

If you combine lot's of return data types, for example `string[]` and `int[]`, or the list of types cannot be estimated in advance, return type `mixed[]` is better, I think to myself.

dg:
I still don't understand how differs `array` vs `mixed[]`.

mabar:
Phpstorm does not make difference between between unknown items type and mixed items type.
Using `mixed[]` is currently useful only if coding standard enforce its usage - it ensures type of items is intentional, not just forgotten.

PR #165, opened 2019-07-09
Fix getter getQueryParameters() to return right property.

janbarasek:
- bug fix
- BC break? yes

Fixed get property.


janbarasek:
Oh, now I see!

Getter `getQueryParameters()` returns `$query` property, but `getQuery()` returns serialized query. I think it's so WTF to debug.

dg:
`getQueryString()` would be better, but it is a tax for gradual development.

PR #168, opened 2019-09-06
Wrong referer annotation propeprty type

lukaspijak:
- bug fix 
- BC break? no

You have wrong type on referer property.


dg:
Thanks

PR #170, opened 2019-10-16
Http response return type

### LINKED ISSUE #169: Nette bridge - service http.response requires return type is Nette\Http\Response instead of Nette\Http\IResponse
Version: 3.0.2

### Bug Description
I am using symfony\browserkit component in integration tests with nette application, the problem is return type from DI container for service `http.response` is specific class instead of interface, which prevents replacing it with own implementation

Here is the response source code: https://github.com/NBrowserKit/NBrowserKit/blob/v2.0.0/src/NetteResponseProxy.php

```php
/** @var Nette\Http\IResponse $response */

$container->removeService('httpResponse');
$container->addService('httpResponse', $response);
```

and i receive
```
Nette\InvalidArgumentException : Service 'http.response' must be instance of Nette\Http\Response, NBrowserKit\Response given.
 /home/users/klarka/wc/17/vendor/nette/di/src/DI/Container.php:83
 /home/users/klarka/wc/17/tests/Integration/BrowserClient.php:33
 /home/users/klarka/wc/17/vendor/symfony/browser-kit/Client.php:407
 /home/users/klarka/wc/17/tests/Integration/Cache/Nocache/NoCacheTest.php:36
```

### Steps To Reproduce
Replace response service with `IResponse` implementation but not `Nette\Http\Response`
```php
$container->removeService('httpResponse');
$container->addService('httpResponse', $response);
```

### Expected Behavior
To work :-) 

### Possible Solution
src/Bridges/HttpDI/HttpExtension.php:
```diff
           $response = $builder->addDefinition($this->prefix('response'))
+                ->setType(Nette\Http\IResponse::class)
                 ->setFactory(Nette\Http\Response::class);
```

JanMikes:
Fixes #169

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
 		"nette/di": "^3.0",
 		"nette/tester": "^2.0",
 		"tracy/tracy": "^2.4",
-		"mockery/mockery": "^1.2@dev"
+		"mockery/mockery": "^1.2"
 	},
 	"conflict": {
 		"nette/di": "<3.0.0-stable"
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
 	"require-dev": {
 		"nette/di": "^3.0",
 		"nette/tester": "^2.0",
-		"tracy/tracy": "^2.4"
+		"tracy/tracy": "^2.4",
+		"mockery/mockery": "^1.2@dev"
 	},
 	"conflict": {
 		"nette/di": "<3.0.0-stable"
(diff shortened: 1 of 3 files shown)


JanMikes:
Another solution could be to change how https://github.com/nette/di/blob/master/src/DI/Container.php#L128 works - if the class implements some interface, then return interface instead of specific implementation, but i am not sure if this change would be desired, though it would fix it too.

PR #172, opened 2019-10-31
IRequest::getServerParameters()

### LINKED ISSUE #166: Request - server parameters
I am missing abstraction for `$_SERVER` in request. I would like to have `IRequest->getServerParameters()`. Should I send a PR?

mabar:
- new feature? yes
- BC break? yes

Closes #166

### DIFF
--- a/src/Http/IRequest.php
+++ b/src/Http/IRequest.php
@@ -117,4 +117,10 @@ interface IRequest
 	 * Returns raw content of HTTP request body.
 	 */
 	function getRawBody(): ?string;
+
+	/**
+	 * Returns data related to the incoming request environment,
+	 * typically derived from PHP's $_SERVER superglobal.
+	 */
+	function getServerParameters(): array;
 }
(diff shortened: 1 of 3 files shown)


dg:
Server parameters really are not part of HTTP request…

mabar:
Not all, but many of them are request-specific, like `HTTP_USER_AGENT` and `REQUEST_METHOD`. Initially I created this PR and related issue because PSR-7 include `getServerParams()` too and in `ServerRequestInterface` docs is statement `$_SERVER values represent application state at the time of request` which sounds reasonable to me. But I see your point of view now. Would an addition of specific methods like `getHttpAgent()` be more acceptable?

dg:
Why you dont like getHeader('User-Agent') ?

mabar:
Nevermind, bad example. It was all about variable abstraction. But it does not belong to request as you pointed, so I may just create separate object for that

PR #173, opened 2019-11-17
Extension: Cookie "nette-samesite" should be secured in case of HTTPS.

janbarasek:
- bug fix
- BC break? yes

I think in case of configure option `cookieSecure === true` the control cookie `nette-samesite` should be secured too.

Currently behavior is like:

image

You can see the cookie `nette-samesite` is not secured and some tests report security issue.

My changes make report more better:

image

But this security policy must be active only in case of `https`.

Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ php:
     - 7.1
     - 7.2
     - 7.3
+    - 7.4snapshot
 
 env:
     - PHP_BIN=php
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
 	"minimum-stability": "dev",
 	"extra": {
 		"branch-alias": {
-			"dev-master": "3.0-dev"
+			"dev-master": "4.0-dev"
 		}
 	}
 }
--- a/src/Bridges/HttpDI/HttpExtension.php
+++ b/src/Bridges/HttpDI/HttpExtension.php
@@ -123,7 +123,7 @@ class HttpExtension extends Nette\DI\CompilerExtension
 			}
 		}
 
-		$code[] = Helpers::formatArgs('$response->setCookie(...?);', [['nette-samesite', '1', 0, '/', null, null, true, 'Strict']]);
+		$code[] = Helpers::formatArgs('$response->setCookie(...?);', [['nette-samesite', '1', 0, '/', null, $config->cookieSecure === true ? true : null, true, 'Strict']]);
 
 		$initialize->addBody("(function () {\n\t" . implode("\n\t", $code) . "\n})();");
 	}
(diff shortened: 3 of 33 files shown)


dg:
I can't simulate this issue.  Argument `null` means the same as `$cookieSecure`

martinolmr:
@dg 

Argument `null` is same as `$cookieSecure`, but in function `setCookie()` (Response.php) property `$this->cookieSecure` is always `false`. This issue is only for 'nette-samesite' cookie, other cookies are sended with secure flag correctly.

**For simulate:**

- in neon set cookieSecure on true
- in Http/Response.php dump `$options` before setcookie

dg:
> Argument `null` is same as `$cookieSecure`, but in function `setCookie()` (Response.php) property `$this->cookieSecure` is always `false`. This issue is only for 'nette-samesite' cookie, other cookies are sended with secure flag correctly.

Can you investigate why it is false, when it should be the same as configuration value?

janbarasek:
@dg I already know where the problem is.

@martinolmr is trying to set the value for the `session` extension and `http` extension, but correctly must define in neon separately.

I suggest that the parameter `cookieSecure` be shared for both. That's why you couldn't simulate it.

janbarasek:
By the way in documentation is example like:

```yaml
session:
       cookieSecure: true
```

but this syntax set secured cookie only for sessions, but not for initial `nette-samesite`. It can be confusing for beginners.

martinolmr:
@dg 

Problem is here (maybe this issue solved):

In nette doc (neon) is:

```
session:
       cookieSecure: true
```

But this not working. Right version is:

```
http:
      cookieSecure: true
```
---

HttpExtension line 65 - loading `http` section from neon not `session`:

```
if ($config->cookieSecure !== null) {
        $value = $config->cookieSecure === 'auto'
		? $builder::literal('$this->getService(?)->isSecured()', [$this->prefix('request')])
		: $config->cookieSecure;
	$response->addSetup('$cookieSecure', [$value]);
}
```
---

Maybe support both version for use and create special page in nette doc. My working neon configuration for secure website:

```
http:
	headers:
		Strict-Transport-Security: 'max-age=15768000'
		Content-Security-Policy: "script-src 'self'
		X-XSS-Protection: '1; mode=block'
		X-Content-Type-Options: nosniff
		Referrer-Policy: 'no-referrer, strict-origin-when-cross-origin'
	cookieSecure: true

session:
    cookieSamesite: Strict
    cookieSecure: true
```

Website secure tester:

https://observatory.mozilla.org

dg:
Cookie `nette-samesite` has nothing to do with the session, so it is related to `http: cookieSecure: true`. 

Actually, it's usually better to set `http: cookie*` instead of `session: cookie*`.

PR #179, opened 2020-05-13
FileUpload::getSanitizedName: Remove redundant minus before dot.

janbarasek:
- new feature
- BC break? yes

In the case of the name `logo 2020+.pdf` old function return `logo-2020-.pdf`. I think more better behavior is simply `logo-2020.pdf` (without minus before dot separator).

Thanks.


dg:
thanks

PR #184, opened 2020-10-11
Fix TypeError if getimagesize() returns false

### LINKED ISSUE #183: Type error in FileUpload->getImageSize()
Version: 3.0.4

### Bug Description
`$file->getImageSize()` returns `TypeError` for the file is not an image. PHP function `getimagesize()` returns `array|false`, not `array|null`. 
See https://github.com/nette/http/blob/master/src/Http/FileUpload.php#L191

### Steps To Reproduce
Upload a file that is not image and ask for `$file->getImageSize()`.

### Expected Behavior
No `TypeError`.

### Possible Solution
Retype result of PHP function `getimagesize()` from `false` to `null`.

dada-amater:
- bug fix #183 
- BC break? no

### DIFF
--- a/src/Http/FileUpload.php
+++ b/src/Http/FileUpload.php
@@ -188,7 +188,8 @@ final class FileUpload
 	 */
 	public function getImageSize(): ?array
 	{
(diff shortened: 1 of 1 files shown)


dg:
Thanks

PR #185, opened 2020-10-23
Lets use dns instead of IP in proxy settings

Ewwe:
Hi, when using nette on reverse proxy i found out when i want to use docker proxy i have to update 

http:
      proxy: [ip]

on every environment i use, instead of using docker created dns after service name in compose.  So this QoL edit let me use dns instead of IP


mabar:
This solution suffer from same problems as nette/bootstrap#54 and nette/tracy#297, it could not be trusted in non-docker environments.

Maybe an environment variable like `DOCKER=true` or `TRUSTED_ENV=true` could be introduced to allow this behavior when it's save to use.

Also DNS resolution can be very slow and add seconds to application response times.

Ewwe:
OK so i edit HttpExtension for accepting new config option with default value false which determines that you are using docker, so depends on this option the setProxy method is affected

Ewwe:
Bude se něco dít nebo ne a prostě zadny nette v dockeru ?

PR #188, opened 2020-11-14
ScriptUrl: Fix class extendability (issue #187)

### LINKED ISSUE #187: Unable to extend UrlScript
Version: 3.0.5
PHP: 7.3.24

### Bug Description
`UrlScript` is not work correctly when is extended. The [method `withPath()`](https://github.com/nette/http/blob/v3.0.5/src/Http/UrlScript.php#L53-L58) is calling themself recursive in infinite loop and crash with error:
```
Maximum function nesting level of '256' reached, aborting!
```

The reason is the callback `parent::withPath` in `call_user_func()` function is always called in top of descendants hierarchy context.

That's mean:
```
class CustomUrl extends UrlScript      parent::withPath()   call_user_func('parent::withPath')
                                         |                                    |
        +--------------------------------+------------------------------------+
        |  +-------------------------------------------------------------------------------------+
        v  v                                                                                     |
class UrlScript extends UrlImmutable   parent::withPath()   call_user_func('parent::withPath')   |
                                         |                                    |                  |       
           +-----------------------------+                                    +------------------+
           |
           v
class UrlImmutable
```

### Steps To Reproduce
```php
class CustomUrl extends \Nette\Http\UrlScript {}

$url = new CustomUrl('http://example.com')'
$url->withPath('foo'); // crash
```

### Expected Behavior
Don't crash ;-)

### Possible Solution
Only reason why is `UrlScript` using `call_user_func()` function to call parent method through is calling in cloned object context.

`UrlScript` is inherits from `UrlImmuitable`, `UrlImmuitable::withPath()` is always create clone too. When calls `UrlScript::withPath()` it's clone class twice.

~~Possible Solution is stop cloning object in `UrlScript` scope and rely it to `UrlImmuitable`.~~

~~I suggest to change [method `withPath()`](https://github.com/nette/http/blob/v3.0.5/src/Http/UrlScript.php#L53-L58) like:~~
```php
public function withPath(string $path, string $scriptPath = '')
{
    $dolly = parent::withPath($path);
    $dolly->scriptPath = $scriptPath;
    return $dolly;
}
```

~~Is it my considerations right?~~ No, isn't, see comment below.

jakubboucek:
- bug fix: #187 
- BC break? no

`UrlScript` is not work correctly when is extended. The [method `withPath()`](https://github.com/nette/http/blob/v3.0.5/src/Http/UrlScript.php#L53-L58) is calling themself recursive in infinite loop and crash with error:
```
Maximum function nesting level of '256' reached, aborting!
```

The reason is the callback `parent::withPath` in `call_user_func()` function is always called in top of descendants hierarchy context (see more at issue #187).

PR is changing relative reference to parent class to absolute. Here is no able to use only `parent` because method is called in another instance than `$this` (cloned one).

Tests:
- All passed.
- No new one necessary

Resolve #187

### DIFF
--- a/src/Http/UrlScript.php
+++ b/src/Http/UrlScript.php
@@ -54,7 +54,8 @@ class UrlScript extends UrlImmutable
 	{
 		$dolly = clone $this;
 		$dolly->scriptPath = $scriptPath;
-		return call_user_func([$dolly, 'parent::withPath'], $path);
+		$parent = \Closure::fromCallable([UrlImmutable::class, 'withPath'])->bindTo($dolly);
+		return $parent($path);
 	}
 
 


src/Http/UrlScript.php
```diff
@@ -54,7 +54,8 @@ public function withPath(string $path, string $scriptPath = '')
 	{
 		$dolly = clone $this;
 		$dolly->scriptPath = $scriptPath;
-		return call_user_func([$dolly, 'parent::withPath'], $path);
+		$parent = \Closure::fromCallable([UrlImmutable::class, 'withPath'])->bindTo($dolly);
+		return $parent($path);
```
JanTvrdik, line 58: `Closure::call()` exists since PHP 7, so it should be possible to simplify this to

```php
Closure::fromCallable([UrlImmutable::class, 'withPath'])->call($dolly, $path)
```

jakubboucek: I tried it first. It works only for anonymous functions. 

```
Cannot rebind scope of closure created by ReflectionFunctionAbstract::getClosure()
```

More info: https://bugs.php.net/bug.php?id=74558

dg:
Thanks

Issue #182, opened 2020-06-12
Cookie nette-samesite


nargotik:
Hi, as far i saw this cookie is set on src/Bridges/HttpDI/HttpExtension.php

There is any reason for this cookie to cannot be optional or name changed ?

nargotik:
Still not figure why this is this way .. and why cant the cookie be renamed.

dg:
Why do you want to rename it?

nargotik:
Because some projects don't need to say the visitors what technology they use, it makes part of security to don't pass information to visitors that provides crucial information about the system.

nargotik:
@dg What you think about it ?

nargotik:
Have added a PR with changes needed to add a new config variable: cookieNameStrict

PR: #189

dg:
At the moment, it is difficult to make a clean solution, so I'll implement it in the next bigger version.

However, in nette/http 3.1 the name has changed to `_nss` which is not so noticeable.

PR #189, opened 2020-12-06
Allow users to change cookieNameStrict in configuration

nargotik:
- new feature? 
Security enhancement by leting users choose to change nette-samesite cookie name  

#182 

Allow users to change cookie name is a security enhancement because sometimes there is no need to say site users what technology site is using.

http:
    cookieNameStrict: fancyname-samesite

- BC break? no
- doc PR: nette/docs#???

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ jobs:
         strategy:
             matrix:
                 os: [ubuntu-latest, windows-latest, macOS-latest]
-                php: ['7.1', '7.2', '7.3', '7.4', '8.0']
+                php: ['7.2', '7.3', '7.4', '8.0']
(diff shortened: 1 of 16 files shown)


dg:
At the moment, it is difficult to make a clean solution, so I'll implement it in the next bigger version.

However, in nette/http 3.1 the name has changed to `_nss` which is not so noticeable.

nargotik:
@dg Ty for the reply, I think that having nette in cookie name is not bad, the bad thing is dont give the dev the flexibility of changing it just because of security.

Normally is a good behavior to dont pass all informations about what is running the system to a possible attacker, even if is _nss it gaves a clue to the possible attacker that is using a nette/http > 3.1

dg:
I understand that, I am leaving the issue open.

PR #190, opened 2020-12-25
Url: IDN domain automatically convert to UTF-8.

janbarasek:
- bug fix
- BC break? yes

By documentation (more info: https://www.php.net/manual/en/function.idn-to-utf8.php) the domain `xn--tst-qla.de` is same as `täst.de`, but Nette Url claims that it is a different URLs:

```php
$url = new Url('https://täst.de:443/cart');
$url2 = new Url('https://xn--tst-qla.de/cart');

$url->isEqual($url2); // false
```

This modification adds automatic conversion of IDN domains to UTF-8 format (if an extension is available). Prior to conversion, the use of IDNs is detected to maintain maximum performance.

If a particular PHP does not have an extension, it can be replaced by the `symfony/polyfill-intl-idn` package.

Thanks.


dg:
Thanks

PR #192, opened 2021-01-14
SessionSection: __get() returns reference (#191)

### LINKED ISSUE #191: Creating array property in SessionSection has no effect
Version: 3.1.0

### Bug Description
When I create sessionSection and save an array property to it with one element, it does not have any effect.

### Steps To Reproduce
```
$cart = $this->session->getSection('cart');			
$cart->items[0] = 0;
echo $cart->items[0];
```
I get notice `Trying to access array offset on value of type null`.

### Expected Behavior
To get `0`

redeyecz:
- bug fix #191
- BC break? no

 Session section __get() should return reference

### DIFF
--- a/src/Http/SessionSection.php
+++ b/src/Http/SessionSection.php
@@ -64,7 +64,7 @@ class SessionSection implements \IteratorAggregate, \ArrayAccess
 	 */
 	public function &__get(string $name)
 	{
-		$data = $this->getData(false);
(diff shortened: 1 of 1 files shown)


dg:
Thaks

janlanger:
@dg This causes nasty bugs, could we get a bugfix release please 🙏

PR #193, opened 2021-01-20
Url: Use more better deprecated message.

janbarasek:
Reference: https://github.com/nette/http/commit/f428947907bd5b6d4d681ece666dc50a4b51bd13

- new feature
- BC break? no

Just info for safe refactoring.


dg:
Thanks

PR #194, opened 2021-03-01
Request: In case of CLI method should be CLI too.

janbarasek:
- new feature
- BC break? yes

In the case of obtaining a service for an Http Request in CLI mode, it does not make sense for the call method to be GET when no request exists.

I think Request should either throw an exception that cannot be used in a CLI context, or it should return a special CLI method.

Current behavior looks illogical:

Screenshot from 2021-03-01 10-06-40

Thanks.

### DIFF
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -77,7 +77,7 @@ class Request implements IRequest
 		$this->files = (array) $files;
 		$this->cookies = (array) $cookies;
 		$this->headers = array_change_key_case((array) $headers, CASE_LOWER);
-		$this->method = $method ?: 'GET';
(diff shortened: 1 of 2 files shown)


dg:
HTTP method CLI doesn't exists

janbarasek:
I don't know how to better represent this (if at all).

When obtaining the current Request from the DI, it should be possible to verify that the Request was created artificially in CLI mode and is not a real request. In that case, I can't rely on a number of things, such as URLs.

What do you think?

milo:
In a strict way - an HTTP request can be automatically created only via HTTP protocol so exception is more accurate. In practice - I'm not sure.

PR #195, opened 2021-03-26
FileUpload: fix getters calls on empty FileUpload

EdaCZ:
- bug fix: yes
- BC break: no

------------------

This happened to me recently: I forgot to mark form item with file upload as mandatory, so users were able to send form without any file. But on backend I have complex system of validators, which want to validate name, size etc. Validators are build using composition and each of them do just one thing (for example one of them just calls `getName` (newly `getSanitizedName` / `getUntrustedName`) on received `FileUpload` and checks its name...). It's not really handy to repeat in each of them condition with `$file->isOk()`... But if I didn't have it there, it would crash on the error: `Return value of Nette\Http\FileUpload::getName() must be of the type string, null returned` (in PHP 7.3), because properties don't have default values and `null` is not compatible with typehint `string`...

I believe that getters should be callable without an error even on empty file upload, to make behavior of our `FileUpload` more predictable and consistent. 

I tried to add failing test and fix it. 

I hope it's ok for you, just let me know if some changes needed.

Thanks.


dg:
fixed

PR #198, opened 2021-04-24
Url: Host should end with dot which should be removed.

janbarasek:
- new feature
- BC break? yes

Domain `https://google.com./abcd` (host ending with dot) is valid URL, but it is same as `https://google.com/abcd`.


dg:
So canonicalize() should be fixed, not constructor.

janbarasek:
@dg Moved. I thought canonize() should be deprecated (from the last version).

dg:
Great. Probably the same applies to isEqual(). Can you also add test?

PR #199, opened 2021-04-24
Url: Add PhpStan typehints to array properties.

janbarasek:
- new feature
- BC break? no

Just improvement.


JanTvrdik:
How is `array<string|int, mixed>` different from just `array`?

janbarasek:
It means the array definition has not been forgoten.

dg:
Better is forgotten than wrong.

PR #210, opened 2022-02-17
Request: Fix Referrer getter for malformed data + test

jakubboucek:
- bug fix
- BC break? no

## What problem is PR solving
Method `\Nette\Http\Request::getReferer()` throws `LogicException` when is called HTTP Request with malformed `Referer` header.

```shell
curl -v -H 'Referer://///' https://example.com/
```

```php
$httpRequest->getReferer();
// throws InvalidArgumentException: Malformed or unsupported URI '/////'.
```

It's wrong, because:

- Exception breaks Application stability according to 🔴 User's input,
- Exception is `Logic` type, but origin of error is only invalid user's input – developer here can't to prevent throw exception,
- Exception is throws too late, when developer call getter – until then it is Request object at incostistence,
- Developer does not except exception here (but be aware to just add `@throws`, it doesn't enough),
- Because Referer is nullable and because most usage of Referer value is logging of processed request, I suppose exception throwing is non-sense here. 

Normal use of Referer value is fo logging:

```php
$logger->logEvent($result, $id, (string)$request->url, $request->method, (string)$request->referer);
```

We can't to wrap whole row to `try/catch`, because it cause to skip log event. Developer must to separate getter call:

```php
try {
    $referer = (string)$request->referer;
}
catch(Nette\InvalidArgumentException $e) {
    $referer = null;
}

$logger->logEvent($result, $id, (string)$request->url, $request->method, $referer);
```

## What PR suggest

PR is suggesting to transmute Exception to Notice, return `null` and never throw Exception. 

Developer always can to read raw header by access through `$request->getHeader('Referer');`

## Security

Potentially this bug can be abused to intentionally bypassing log of user's activity, because in most cases is `Referer` value used for auditing done operations.

### DIFF
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -191,9 +191,16 @@ class Request implements IRequest
 	 */
 	public function getReferer(): ?UrlImmutable
 	{
-		return isset($this->headers['referer'])
-			? new UrlImmutable($this->headers['referer'])
-			: null;
+		if (!isset($this->headers['referer'])) {
+			return null;
+		}
+
+		try {
+			return new UrlImmutable($this->headers['referer']);
+		} catch (Nette\InvalidArgumentException $e) {
+			trigger_error("Unable to parse Malformed Referer URI: {$this->headers['referer']}");
+			return null;
+		}
 	}
 
 
--- a/tests/Http/Request.referer.phpt
+++ b/tests/Http/Request.referer.phpt
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * Test: Nette\Http\Request headers.
+ */
+
+declare(strict_types=1);
+
+use Nette\Http;
+use Tester\Assert;
+
+
+require __DIR__ . '/../bootstrap.php';
+
+
+test('', function () {
+	$request = new Http\Request(new Http\UrlScript);
+
+	Assert::null($request->getReferer());
+});
+
+
+test('', function () {
+	$request = new Http\Request(new Http\UrlScript, null, null, null, [
+		'referer' => 'http://nette.org:8080/file.php?q=search',
+	]);
+
+	Assert::same('http://nette.org:8080/file.php?q=search', $request->getReferer()->getAbsoluteUrl());
+});
+
+
+test('', function () {
+	$request = new Http\Request(new Http\UrlScript, null, null, null, [
+		'referer' => '/////',
+	]);
+
+	Assert::error(function () use ($request) {
+		Assert::null($request->getReferer());
+	}, E_USER_NOTICE, 'Unable to parse Malformed Referer URI: /////');
+});


dg:
I want to remove this method completely because it makes no sense. Who uses the Referer these days?

jakubboucek:
True true, `@deprecated` annotation looks great solution.

diegosardina:
Guys, the referer is useful to see where the user came from. Google, Facebook, LinkedIn, etc set the referer

diegosardina:
Even duckduckgo, that is privacy-oriented, sets the referer:

`... [17/Feb/2022:20:02:02 +0100] "GET / HTTP/1.1" 301 242 "https://duckduckgo.com/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.102 Safari/537.36"`

Some customers (unfortunately...) want their statistics in his blog administration page outside Google Analytics (that also use the referer by the way)

dg:
Sure, there is a `getHeader('Referer')` method. But no extra method is needed for this header.

diegosardina:
I think it's handy to have that method, since by definition the referer is an url, I'm going to save an instance of ?UrlImmutable (instead of ?string) in a method class

dg:
> Even duckduckgo, that is privacy-oriented, sets the referer:

The referer doesn't send the web, your browser does. But a lot of extensions and proxies block it nowadays. That's why the `Origin` header was created as a successor.

diegosardina:
> The referer doesn't send the web, your browser does.

Of course I know :-) I used "they sets" in an improper way, since they instruct the browser via the referrer-policy, I expected duckduckgo to set "no-referrer" (but it seems they want to appear in web analytics like Google Analytics).

> But a lot of extensions and proxies block it nowadays. That's why the `Origin` header was created as a successor.

The use of proxies or extensions are less than 1% today. By the way I use ublock (with tracking protection, etc..) and my browser sends the referer. Maybe they block it if the policy is different than "origin" (that behaves like Origin header I guess), but I'm lazy to test :-)

diegosardina:
Uhm, there is an official document that says "referer" is deprecated?

What I found [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/referrer) `Content-Security-Policy: referrer` is deprecated (in favour of `Referrer-Policy`) but not the referer itself.

While origin is similar to the referer, currently has different use cases and the browser doesn't send it for normal navigation

dg:
Okay, let me rephrase that. The Referer header is now an old-fashioned unreliable thing, which therefore doesn't deserve to have its own method in the fairly simple Http\Request interface. And if the current method is not well usable, I can in good conscience give it away and leave the solution to userland.

diegosardina:
Ok I won't insist, let's deprecate it.

However I would remove `in favor of the getOrigin()`, because origin is (currently?) used in CORS requests and has different uses than referer

PR #211, opened 2022-03-01
Request: Move Basic Auth credential from Url to Request due to prevent leak it

jakubboucek:
- new feature
- BC break? yes
- doc PR: nette/docs#932

Currently is too risky use `Url` object from from `Http\Request` created from global, because it can simply cause to leak user's credentials:

```php
echo $httpRequest->getUrl();

// https://example.com@username:password/index.php
```

The `Url` object already provide method `–>withoutUserInfo()` to remove user credential, but when programer forgot to use that, the credentials can leaks.

This PR ir removing user's credentials from `Url` object and move it to `Http\Request` directly.

It's based on czech discussion at Nette Forum: https://forum.nette.org/cs/33127-http-basic-login-http-url-nette-3-security-issue

I know, it's too naive PR, here is missing any way to notify developer the `Url->getUser()` no more get user from current request. This PR sent as initiation of change insecurity.

### DIFF
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -52,6 +52,10 @@ class Request implements IRequest
 	/** @var ?callable */
 	private $rawBodyCallback;
 
+	private ?string $user;
+
+	private ?string $password;
+
 
 	public function __construct(
 		UrlScript $url,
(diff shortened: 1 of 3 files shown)


carloscz:
this should be released as new major to comply with semver :/

dg:
@carloscz There were security reasons.

PR #212, opened 2022-04-12
RequestFactory: support full_path

jiripudil:
- new feature
- BC break? no
- doc PR: nette/docs#937

As a follow-up to #207, this adds support for correctly transforming directory uploads into FileUpload objects to RequestFactory.


dg:
Thanks

PR #220, opened 2023-02-24
RequestFactory: Optimize match proxy CIDR with remote IP

jakubboucek:
- BC break? no
- doc PR: not relevant

We are using CloudFlare reverse proxy to run production. [Cloudflare has proxy servers at 22 IP ranges](https://www.cloudflare.com/ips/). Nette `RequestFactory` is on **every** request must match user's IP with 22 CIDR IP masks.

https://github.com/nette/http/blob/4ca7b03d5726cf489f672a0e2a13178318a3d201/src/Http/RequestFactory.php#L283

⇧ this current construction is consuming unnecessarily much CPU/time, because:
- PHP is uneffective with call callback fuctions on loop ([read more about it](https://margaras.k47.cz/@k47/statuses/01GPZSGNYZ2NZWF354FAZHP19R) from @kaja47)
- Loop traversing is not lazy – it vainly continue in loop even have match.

Currently it's consuming about ~0.8 ms, that's ~1/10 of process whole request.

This PR suggest a little uglier but much faster constrictions which is up to:
- 20 × faster when matched (~0.04 ms),
- 4 × faster when unmatched (~0.2 ms).

No functionality changes.

### DIFF
--- a/src/Http/RequestFactory.php
+++ b/src/Http/RequestFactory.php
@@ -280,7 +280,14 @@ class RequestFactory
 			: null;
 
 		// use real client address and host if trusted proxy is used
-		$usingTrustedProxy = $remoteAddr && array_filter($this->proxies, fn(string $proxy): bool => Helpers::ipMatch($remoteAddr, $proxy));
+		$usingTrustedProxy = false;
+		if ($remoteAddr) {
+			foreach ($this->proxies as $proxy) {
+				if ($usingTrustedProxy = Helpers::ipMatch($remoteAddr, $proxy)) {
+					break;
+				}
+			}
+		}
 		if ($usingTrustedProxy) {
 			empty($_SERVER['HTTP_FORWARDED'])
 				? $this->useNonstandardProxy($url, $remoteAddr, $remoteHost)


dg:
It wouldn't be possible to use [Arrays::some](https://doc.nette.org/en/utils/arrays#toc-some) here, would it?

jakubboucek:
@dg Is here static callback any benefit even optically removed one level nested block? 

As I said above, the callback only is slowing whole match up to 4× (independently to lazy loop). I see removing of callback here as necessary optimization.

dg:
The point is that callbacks are used in a huge number of different places. Nette has no doctrine of "not using callbacks" for performance reasons. It's possible to establish such a doctrine if it leads to noticeably better performance, for example in the past Nette had a [don't use a ternary operator](https://phpfashion.com/php-puvab-optimalizace-rychlosti) doctrine because it was extremely slow.

So I don't see why in this particular case to do an optimization that saves 0.00006s per request.

jakubboucek:
I am not trying to introduce new general doctrines for Nette. I like callbacks. I like ternary operator. I don't like stupid „academic“ optimizations.

I believe that in this case it is a specific situation where usage of callback has no benefits but it's slowing down goodly optimized app about 1/10 of time for each request. It just for 22 IP ranges for Cloudflare. What about if somebody will use AWS CloudFront where is [about 150 IP ranges to match](https://d7uri8nf7uskq.cloudfront.net/tools/list-cloudfront-ips).

### Why I mean this is specific case:

- Callback hasn't here any benefit except habit.
- It's executed for every single request, including all redirects, ajax, api, etc. 
- Here is not effective way to cache results.
- It's part of code which is very important for app Security. Is important to preventively make no any barrier which can user demotivate to use it – some company management they tend to prioritize performance over safety.

PR #222, opened 2023-03-09
RequestFactory: fix X-Forwarded-Host mixup with remote host

JanTvrdik:
When reading https://github.com/nette/http/pull/221, I noticed `X-Forwarded-Host` is incorrectly used for remove host.

**X-Forwarded-Host is in no way related to remote host**.
X-Forwarded-Host contain value of `Host` header that client sent to the proxy.
Remote host is related to remote addr (client ip).


dg:
Thanks

PR #224, opened 2023-05-15
AVIF image FileUpload support

jhabaj:
- new feature
- BC break? no
- doc PR: (https://github.com/nette/docs/pull/1011)

Added AVIF image support for FileUpload


dg:
Thanks

PR #229, opened 2023-11-13
Unify and fix host parsing in RequestFactory

Izolex:
- bug fix
- BC break? yes

The implementation of `useForwardedProxy` removes square brackets around the host filled into the resulting URL, which conflicts with [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2). The specifications state that an IPv6 host should always be encapsulated in square brackets - correctly implemented in the `getServer` method. This PR introduces a new method `parseHost`, which is now used in both of these methods to unify the implementation.


dg:
Thanks

PR #230, opened 2023-11-15
RequestFactory: extract port from x-forwarded-host

Izolex:
The unification of host and port parsing was added in #229. Initially, I was uncertain whether the distinct parsing approaches for the `Host` and `X-Forwarded` headers compared to the `X-Forwarded-Host` header were intentional, since the `X-Forwarded-*` headers lack a formal specification. However, it's commonly observed that the `X-Forwarded-Host` header may contain the value of the `Host` header, which may include the port. Currently, if this header includes both the host and the port, the entire string is simply stored in the host field of the resulting URL. The goal of this PR is to parse this header using the `parseHostAndPort` method, to ensure that the port is correctly separated and stored into the port field.

Since this PR uses the method added in #229 it's based on the commit that was merged into the master branch.


dg:
Thanks

PR #235, opened 2024-03-06
Initialize properties of FileUpload class

finwe:
- bug fix & new feature, issue #234 
- BC break possible
- doc PR: probably not needed

As mentioned in https://github.com/nette/http/issues/234, this tries to fix potential inconsistencies and errors with "empty" uploaded file fields.

BC break is not probable, unless someone expects the cast to string to return an error when the tmpName is not initialized.


dg:
Thanks

PR #237, opened 2024-03-22
Add DateTimeInterface to setCookie $expire param

### LINKED ISSUE #233: setCookie() expire type issue
Version: 3.3.0

### Bug Description
Hello @dg,

in interface `Nette\Http\IResponse` you add to `setCookie` method to expiration parameter union type `?int`. But in documentation you can use string date representation. For example in `Nette\Bridges\SecurityHttp\CookieStorage` has property $cookieExpiration union type `?string`, but on line 51 you use `?string ` in `setCookie`, it can cause type error.

Will it be possible to specify expiration using `string|int|null` in the future, or will everything be refactored to `?int`?

spaze:
Adds `string` and `DateTimeInterface` back to `setCookie` `$expire` param .

This is a fix for an undocumented (and possibly unwanted) BC break in 3.3.0.
In version 3.2.x and older, this was the type in the docblock.

Also fix the readme which had a wrong param name.

Close #233

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -307,7 +307,7 @@ Sets the expiration of the HTTP document using the `Cache-Control` and `Expires`
 $httpResponse->setExpiration('1 hour');
 ```
 
-setCookie(string $name, string $value, $time, string $path = null, string $domain = null, bool $secure = null, bool $httpOnly = null, string $sameSite = null)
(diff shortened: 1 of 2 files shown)


dg:
Thanks

Issue #208, opened 2021-10-28
Notices are ignored on session_start()


forrest79:
Version: 3.1.3

### Bug Description
We're using (phpsession)[https://github.com/phpredis/phpredis/] to handle sessions in Redis with a default locking system. There is a notice raised when a lock can't be acquired (https://github.com/phpredis/phpredis/blob/c588f0308a4d49368a436e660b4a29a0579f802d/redis_session.c#L768). The `session_start()` returns `true` even with this notice. Up to version 3.1.2 an exception was thrown, when the lock can't be acquired. Since version 3.1.3 the notice is ignored, session is started and at the end of the request, there is a strange warning logged in Tracy `Warning: Unknown: Failed to write session data (redis). Please verify that the current setting of session.save_path is correct (tcp://csfd-service:6380?timeout=3&database=1)` with no stack trace.

### Steps To Reproduce
Use PHP extension phpredis with similar settings:

```yaml
php:
  redis.session.locking_enabled: 1
  redis.session.lock_wait_time: 3000 # us
  redis.session.lock_retries: 1

session:
  autoStart: false
  save_handler: redis
  save_path: 'tcp://%caching.redis.host%:%caching.redis.ports.session%?timeout=3&database=%caching.redis.databases.session%'
```

Start session in PHP:

```php
$container->getByType(Nette\Http\Session::class)->start();
```

Run two concurrent requests. The second one should receive timeout notice. In 3.1.2 version, we have got an exception, in 3.1.3 is notice ignored.

### Expected Behavior
The exception is raised on the session starts if there is a notice raised in `session_start()`.

### Possible Solution
Return back `Nette\Utils\Callback::invokeSafe` for `session_start` for new minor versions. And possibly add a setting whether a user wants to ignore NOTICES/WARNINGS on session start in a new major version?

PR #240, opened 2024-09-17
Session: keep exceptions chain on session_start

forrest79:
- new feature
- BC break? no

There was an issue #208 - phpredis session driver was raising a notice, when session lock can't be obtained. Nette Session was correctly converting this notice to an exception. The strange was, that even the notice was raised, `session_start` returned `true` (it was phpredis problem).

Now, the new production version of phpredis is out, and the behavior is a little bit different. If a lock can't be obtained, instead of notice, the warning (instead of a notice) is raised and the `session_start` returned correctly `false`.

The problem is, I need to know the reason why the session wasn't started (when redis is death, it's a problem, when session lock can't be obtained, it's probably OK - user is probably doing many parallel requests). `Nette\Utils\Callback::invokeSafe` around the `session_start` method correctly converts warnings to exceptions, but now, there are two exceptions:
- the first is from the phpredis - something like `Failed to acquire session lock`
- the second is from the PHP itself `Failed to read session data: redis (path: tcp://server:6379?timeout=3&database=1&prefix=session:v1:)`

And the first one is overwritten with the second one, that is thrown to the app. But I need to know also the first one :-)

So my first idea is to set correct `previous` exception property and in the app, I have the complete exceptions chain, that I can read and get the correct session problem.

### DIFF
--- a/src/Http/Session.php
+++ b/src/Http/Session.php
@@ -100,7 +100,7 @@ class Session
 				'session_start',
 				[['read_and_close' => $this->readAndClose]],
 				function (string $message) use (&$e): void {
-					$e = new Nette\InvalidStateException($message);
+					$e = new Nette\InvalidStateException($message, previous: $e);
 				},
 			);
 		} catch (\Throwable $e) {


dg:
thanks

